PR #456, opened 2021-08-09
Theming linechart

carysmills:
### What problem is this PR solving?
Resolves https://github.com/Shopify/polaris-viz/issues/405
Adds theming to the line chart

Some notes about changes/decisions I made and a few questions (have also left some comments where relevant in the code):
- I added the dotted line style option to the sparkline so we can share the type. Any objections?
- I made hasSpline part of the theme config. Does that seem like the right place, or should it be configured as a prop still? Just wondering if there are some charts where the type of data doesn't lend itself to splines. This would still be possible with an additional theme, though.
- I added `sparkArea` as well as `area` because they work differently for the Sparkline and Line Chart. The Sparkline accepts a string, null or gradient stop. The line chart just accepts a color and we determine the gradient internally. Changing that on this PR felt out of scope, but let me know if you disagree.
- hiding the xAxis is something that is only possible for the line chart currently. Right now, I have added it to the xAxis theme object. Perhaps it belongs on the line chart props instead, since it isn't applicable to the other charts. Thoughts?
- Since the crosshair color option accepts rgba values, I have removed the opacity key/value that was previously accessible via props

### Reviewers’ :tophat: instructions
Make sure all configuration is still possible for the line chart and sparkline as before

### Before merging

- [ ] Check your changes on a variety of browsers and devices.

- [ ] Update the Changelog.

- [x] Update relevant documentation.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -38,6 +38,13 @@ export const decorators = [
               padding: '20px',
             },
           },
+          NoSpline: {
+            line: {hasSpline: false},
+          },
+          NoxAxisLabels: {
+            xAxis: {hide: true},
+          },
+          NoOverflow: {grid: {horizontalOverflow: false}},
           Positive: {
             bar: {
               color: 'rgb(0, 164, 124)',
--- a/documentation/code/LineChartDemo.tsx
+++ b/documentation/code/LineChartDemo.tsx
@@ -122,7 +122,6 @@ export function LineChartDemo() {
           xAxisOptions={{
             xAxisLabels,
             labelFormatter: formatXAxisLabel,
-            showTicks: true,
           }}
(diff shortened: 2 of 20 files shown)


cispl-shaswatad:
Hi, I am looking for directions to have multiple lines projected on a single Line Chart with different lines having `line.hasSpline` set to true OR false as required.
 
Actually, I am trying to project a regression line in a Line Chart against the main data. However, due to minor decimal approximation issues, the regression line is appearing slightly wavy, significant enough to catch attention when `line.hasSpline: true` is set in the theme. The line appears visually straight with `line.hasSpline: false` essentially solving my problem. However, with `line.hasSpline: false`, the main data line also loses the splines and looks kind of harsh!

I want to have the main line with the Splines and the regression line without.

PR #1359, opened 2022-08-11
Clamp large rounded corners to bar dimensions

envex:

If the combined radius of the border was larger than either the height or width, a rect would draw all funky.

image

Our previous fix was to just remove the rounded borders and draw a basic rectangle.

image

Now that we allow users to set their own border radius via the theme, I think it's better is we still render a rounded corner, but clamp it to the largest radius we can use without breaking the rect.

image
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-oginxkncbh.chromatic.com/?path=/story/chromatic-utilities-getroundedrectpath--horizontal

### Before merging

### DIFF
--- a/packages/polaris-viz-core/src/utilities/stories/getRoundedRectPath.chromatic.stories.tsx
+++ b/packages/polaris-viz-core/src/utilities/stories/getRoundedRectPath.chromatic.stories.tsx
@@ -1,0 +1,100 @@
+import React from 'react';
+import {storiesOf} from '@storybook/react';
+import {getRoundedRectPath} from '../getRoundedRectPath';
+
+const DIMENSIONS = {
+  horizontal: [
+    {
+      height: 5,
+      width: 100,
+    },
+    {
+      height: 20,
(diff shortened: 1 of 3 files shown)


packages/polaris-viz-core/src/utilities/stories/getRoundedRectPath.chromatic.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div
+        style={{
+          display: 'grid',
+          gap: '20px',
+          gridTemplateColumns: 'repeat(4, 1fr)',
+        }}
+      >
+        {BORDER_RADIUS.map((borderRadius) => {
+          return RADIUS_CASES.map((radiusCase) => {
+            console.log(radiusCase);
```
rita-morozova, line 63: console.log 😃

rita-morozova (approved):
I love this idea!

krystalcampioni (approved):

PR #1361, opened 2022-08-11
Position points correctly when not animated

envex:

When `<StackedAreaChart />` was not animated we were not returning the correct x/y values when hovering over different points.

Resolves https://github.com/Shopify/polaris-viz/issues/1355

**Before**

https://user-images.githubusercontent.com/149873/184239062-8ba3b4f3-d6a5-4ef3-95b7-b51c1b394070.mov

**After**

https://user-images.githubusercontent.com/149873/184238879-02e2c17c-e8bf-4d18-8286-a8433bbc4af7.mov
 
### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/StackedAreaChart/components/Points/Points.tsx
+++ b/packages/polaris-viz/src/components/StackedAreaChart/components/Points/Points.tsx
@@ -61,17 +61,19 @@ export function Points({
 
   return (
     <React.Fragment>
-      {stackedValues.map((_, stackIndex) => {
+      {stackedValues.map((stack, stackIndex) => {
         if (activePointIndex == null) {
           return null;
         }
+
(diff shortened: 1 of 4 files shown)


packages/polaris-viz/src/components/StackedAreaChart/components/Points/Points.tsx
```diff
@@ -61,17 +61,19 @@ export function Points({
 
   return (
     <React.Fragment>
-      {stackedValues.map((_, stackIndex) => {
+      {stackedValues.map((stack, stackIndex) => {
         if (activePointIndex == null) {
           return null;
         }
 
+        const [_, y] = stack[activePointIndex];
```
rita-morozova, line 69: Question: why do we use an underscore in this case?

krystalcampioni (approved):

PR #1367, opened 2022-08-18
Organize storybook stories

envex:

Organized all our stories into single files with shared `Meta` and default data.

Now each stories folder should contain a `meta.tsx` and `data.tsx` file. Each story consumes these defaults so everything is available in a consistent system.

Nothing should be visibly changed, all the stories should still work as they used to.


envex:
@LBWright @krystalcampioni @rita-morozova This is ready for review!

LBWright (approved):
It's a little hard to review, but I really like this change quite a bit.

PR #1380, opened 2022-08-31
Don't render TooltipWrapper when data could be bad

envex:

When a data series has an empty data array, some of the charts will error because the tooltips expect valid data.

```
[{name: 'Empty', data: []}]
```

Instead of checking for bad data, we can not render the `<TooltipWrapper />`.

I also added a warning in the console if the user passes mismatched data series. Currently the chart goes into an error state, but the consumer may not know why.

Resolves https://github.com/Shopify/polaris-viz/issues/1328

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/ChartErrorBoundary/ChartErrorBoundary.tsx
+++ b/packages/polaris-viz/src/components/ChartErrorBoundary/ChartErrorBoundary.tsx
@@ -1,11 +1,19 @@
-import {ChartState, Dimensions} from '@shopify/polaris-viz-core';
+import {
+  ChartState,
+  DataGroup,
+  DataSeries,
+  Dimensions,
+} from '@shopify/polaris-viz-core';
 import React, {ErrorInfo} from 'react';
 
 import {ChartSkeleton} from '../ChartSkeleton';
 import type {SkeletonType} from '../ChartSkeleton';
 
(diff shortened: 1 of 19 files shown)


packages/polaris-viz/src/components/ChartErrorBoundary/ChartErrorBoundary.tsx
```diff
@@ -55,3 +63,22 @@ export class ChartErrorBoundary extends React.Component<
...
+  }
+
+  const firstSetLength = data[0].data.length;
+
+  const hasMismatchedData = data.some(
+    (series) => series.data.length !== firstSetLength,
+  );
+
+  if (hasMismatchedData) {
+    console.warn(
```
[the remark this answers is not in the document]
envex: The data included in the chromatic story is for the `<TooltipWrapper />` issue, not the mismatched data.

pbojinov, line 79:
```suggestion
   // eslint-disable-next-line no-console
    console.warn(
```

pbojinov: ESlint is complaining about the console log

image

LBWright (approved):

Issue #153, opened 2020-10-27
[Multi series bar chart] Replace the use of label in `get-stacked-values`, use another unique value


KyleKilbride:
Currently, the use of `label` as a key in `get-stacked-values` does not work well for non-unique labels when creating the `stacked values` and the `y-scale domain`. 

the issue lives in the call to the stack generator:
```
const barStack = stack()
    .offset(stackOffsetDiverging)
    .keys(series.map(({label}) => label));
```

example of what we don't handle, but would like to:
```
const data: Data[] = [
      data: [1, 2, 3], color: 'primary', label: 'oranges',
      data: [10, 20, 30], color: 'secondary', label: 'oranges',
]
```

PR #1384, opened 2022-09-07
Use unique values for stack() keys

envex:

Use a unique key for stacked values in case the consumer doesn't use a unique name for each data series.

Resolves https://github.com/Shopify/polaris-viz/issues/153

Nothing would crash the charts, but the logic for figuring out which stack needed rounded corners would be wrong.

| Before  | After  |
|---|---|
|image|image|


rita-morozova (approved):
👍

LBWright (approved):

PR #1386, opened 2022-09-07
Use a single options object for useYScale

envex:

In some charts, we have to call `useYScale` twice. Once to get initial values and then a second time once the chart has resized itself based on labels & legends.

We were passing the same data twice, which lends itself to bugs.


rita-morozova (approved):
👏 👏 👏

PR #1388, opened 2022-09-07
Animate vertical bar path

envex:

When adding a dynamic content change to the `VerticalBarChart` I noticed that we were animating the scale from 0 to 1 which meant we would not be able to animate data changes.

This adds the animation on the path so data changes will animate between states.
 
### Before merging


rita-morozova (approved):
👍

PR #1398, opened 2022-10-11
Account for empty tooltip data

chrisdakin:
Sometimes the data passed into tooltip is empty:

image

This is causing the LineChart component to break when mousing further right than the existing data. This PR ensures that the data exists before referencing it.

Closes https://github.com/Shopify/core-issues/issues/45475

### Before merging


chrisdakin:
@envex I added a `CohortComparisonDataSet` story where you can see the chart breaking without the changes in `useGetLongestLabelFromData.ts`, and I added a test in `useGetLongestLabelFromData.test.tsx` that fails without the change. I tried adding a test in `LineChart.test.tsx` but didn't manage to replicate the mousemove behavior necessary to trigger the bug. Do you think the first two changes suffice or do we need the `LineChart` test as well?

envex (approved):
Looks good, I've created a follow issue that we can handle for the empty tooltip.

https://github.com/Shopify/polaris-viz/issues/1400

PR #1406, opened 2022-10-19
Use dist versions of d3 when testing

envex:

Moving from `transformIgnorePatterns` which requires d3 packages to be transpiled.

Transpiling has a speed hit so moving to the built versions of d3 in jest is faster.


pbojinov (approved):
Code looks good. Not sure how to 🎩

PR #1429, opened 2022-11-04
Remove LineSeries.scss from core

envex:

We're having issues bringing in the `polaris-viz-core/styles.css` file in web. Because of this the LineSeries elements were still getting the focus style applied to them.

We can remove it from core and just apply the styles directly in the `polaris-viz` styles.

### Before merging


blboyle:
https://screenshot.click/04-05-xg887-h2k9d.mp4

Yea this looks good.

I will say that the other places that do use the lineseries will have the blue ring around it but I do not think they are used (ComboChart) or at least aren't focusable (SparkLine).

blboyle (approved):

PR #1438, opened 2022-11-16
switch to curveMonotoneX from curveStepRounded for smoothing lines

etlgfx:

Switching from `curveStepRounded` to `curveMonotoneX` across the board for spline / line / area smoothing.

`curveStepRounded` suffers from stair stepping artifacts, that to me communicate a local stagnation in the value being charted, wheras `curveMonotoneX` smoothes the line as a gradual slope which communicates a continuous trend, which I believe more accurately conveys the underlying data we're presenting in timeseries graphs.

no, wanted to start a discussion :)

before: `curveStepRounded`
Screenshot 2022-11-16 at 12 41 42

after: `curveMonotoneX`
Screenshot 2022-11-16 at 12 39 20

before: `curveStepRounded`
Screenshot 2022-11-16 at 12 41 33

after: `curveMonotoneX`
Screenshot 2022-11-16 at 12 41 25

Examples specifically to illustrate the stair stepping effects 🙏🏽 

Both solutions support these important features:

- line does not exceed minima/maxima in the data set
- line directly connects to each data point
- line is flat between two subsequent equal values (monday: 10, tuesday: 10 => horizontal line)
- nice smooth line
 
## Storybook link

All charts for lines & area charts render slightly differently with this change.

### Before merging


adamperron:
https://docs.google.com/document/d/1U44R629BVJVHyeGJSVJv7U4JMYBTEr7xcjjpwqAjUlo/edit#heading=h.yfgopa9sa9c5

adamperron:
Hi Eric, you’re right the doc doesn’t included our explorations on the curvature of the lines. I’ll include it here once I can track it down.

In the meantime, the overall sentiment on this issue is that we agree that the wave artifacts aren’t ideal, but it’s a trade-off we are willing to make over the asymmetry of the bezier curves from `curveMonotoneX`. And since line interpolations are inferred trajectory from point A to B, neither curvature styles are necessarily more accurate, so we're choosing what we feel is the lesser of aesthetic evils.

etlgfx:
Is it this one? https://www.figma.com/file/PyEzqNCr5fvSPQyoD7aC4J/Curve-Step-Rounded?node-id=0%3A1&t=wHGIv5SYonv5bJ2t-0

Agreed that lines imply data that doesn't exist, and points / bars are probably the only 'accurate' representation of the data we're actually surfacing.

Screenshot 2022-12-14 at 15 23 38

This is sort of the main example of the visual noise that I think is just much worse on the `curveStepRounded` vs `curveMonotoneX`, it's a continuously increasing (or decreasing, same stepped artifacts happen) range of intervals, so "should" be represented as a continuously increasing line. The steps imply even more non-existing data in these cases. And now I can't unsee it.

I won't keep banging this drum forever of course, just wanted to make sure the thing I'm trying to improve on is well understood, happy with whatever decision you guys make.

PR #1442, opened 2022-11-23
Rita morozova/change comaprison line

rita-morozova:
### Before merging


packages/polaris-viz-core/src/components/SparkBarSeries/SparkBarSeries.tsx
```diff
@@ -133,7 +131,7 @@ export function SparkBarSeries({
           strokeLinecap="round"
           opacity="0.9"
           strokeDashoffset={strokeDashoffset}
-          strokeDasharray={strokeDasharray}
+          strokeDasharray="0.1 4"
```
rita-morozova, line 134: I'm not sure if this is the best option to hardcode this value but we need to match the same value as in the `LineChart` and the `SparkLineChart`

Screenshot 2022-11-23 at 11 07 13 AM

envex: Can you make this constant we can share between `LineChart` and `SparkLineChart`?

PR #1447, opened 2022-11-30
Re-export ChartState from `@shopify/polaris-viz` as value instead of type

andrewzulaybar:

This PR exports the enum `ChartState` as a value instead of a type. This will allow developers to use `ChartState` when passing in the `state` prop to various charts, without having to install and import it from `@shopify/polaris-viz-core`.

Currently, this error appears when trying to use any of `ChartState.Success`, `ChartState.Loading`, or `ChartState.Error` after importing it from `@shopify/polaris-viz`:
image

Seeing as this is how `ChartState` is exported from [`polaris-viz-core`](https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz-core/src/index.ts#L119), I think it would make sense to match that.

N/A

N/A
 
## Storybook link
[Storybook](https://6062ad4a2d14cd0021539c1b-pxkqyjyvnk.chromatic.com/)

### Before merging

### DIFF
--- a/packages/polaris-viz/src/index.ts
+++ b/packages/polaris-viz/src/index.ts
@@ -51,6 +51,7 @@ export {
   getAverageColor,
   paddingStringToObject,
   removeFalsyValues,
+  ChartState,
   ColorScale,
 } from '@shopify/polaris-viz-core';
 
@@ -62,5 +63,4 @@ export type {
   PartialTheme,
   GradientStop,
   DataPoint,
-  ChartState,
 } from '@shopify/polaris-viz-core';


envex:
I don't think you have to install core as it's own package, you should be able to reference it from core if you have `@shopify/polaris-viz` installed, no?

andrewzulaybar:
> I don't think you have to install core as it's own package, you should be able to reference it from core if you have `@shopify/polaris-viz` installed, no?

Ahh, I see what you mean. I'm able to reference it but it looks like I get a lint error which I think might be coming from `sewing-kit`? 
image

andrewzulaybar:
@envex Oh that's interesting! Must be something specific to the repo I'm working in ([Shopify/fbs](https://github.com/Shopify/fbs)). In this case, I don't think we mind adding in the extra dependency so this is probably not needed. I'll go ahead and close this PR

PR #1461, opened 2023-01-09
RFC: Benchmarks project LineChart

envex:

The Benchmarks project introduces a visual element that visually "groups" lines to create a top/medium/bottom percentile "group".

I have a few ideas on how we can do this, while keeping the components clean and reusable.

**Idea 1**

Export out a new component `<LineChartRelational />` that is a wrapper on the `<LineChart />` component.

Along with idea 2, this allows us to reuse all of the `<LineChart />` logic without muddying up the api with new props.

**Idea 2**

Introduce the concept of "slots" for each chart that would allow us to render new elements in an existing chart.

**Idea 3**

Introduce a `metadata` field our `DataSeries` type. We could type this per chart that uses it which would allow us to put anything in `metadata` that the chart required.

image
 
### Before merging


packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -327,6 +329,8 @@ export function Chart({
           transform={`translate(${chartXPosition},${chartYPosition})`}
           className={styles.Group}
         >
+          {slots?.chart?.({yScale, xScale})}
```
envex, line 332: Idea 2: This is the slot area used to render the `<RelatedArea />` component.

packages/polaris-viz/src/components/LineChartRelational/stories/data.tsx
```diff
@@ -0,0 +1,161 @@
...
+      {value: 791, key: '2020-03-12T12:00:00'},
+      {value: 623, key: '2020-03-13T12:00:00'},
+      {value: 21, key: '2020-03-14T12:00:00'},
+    ],
+    color: '#286A7B',
+    metadata: {
+      relatedIndex: 3,
+      areaColor: '#F0F8FB',
+    },
+    styleOverride: {
```
envex, line 131: image

Based on the Figma design we need to allow the consumer (us) to make changes to individual line styles. 

My idea is to add another property to `DataSeries[]` that would allow us to customize each line specifically instead of having to rely on the theme (or having no customization at all). Right now I've only added the properties I need, but if we like this approach we could apply this to all charts that can be customized.

This isn't set in stone, I'm looking for some feedback on the approach.

cc: @pbojinov @LBWright @nickpresta @philschoefer

envex: Ideally we would set defaults for the conumsers, but enable the ability for them customize each series individually - or globally via the theme like they can today.

PR #1464, opened 2023-01-13
Allow consumers to trigger and listen to active events

envex:

We want users to be able to trigger, and respond to, active series changes.

This is an initial draft to support the Benchmarks project: https://vault.shopify.io/gsd/projects/30453 which is why were only supporting `LineChart` for this version.

If we find that there's a genuine need for this in other charts, we'll add the functionality later.
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-tsnlqfbano.chromatic.com/?path=/story/polaris-viz-playground-externalevents--default

### DIFF
--- a/packages/polaris-viz/src/hooks/useWatchActiveSeries.ts
+++ b/packages/polaris-viz/src/hooks/useWatchActiveSeries.ts
@@ -1,0 +1,24 @@
+import {COLOR_VISION_SINGLE_ITEM} from '@shopify/polaris-viz-core';
+import {useEffect} from 'react';
+
+import type {CustomEventReturn} from '../types';
+
+import {getEventName} from './ColorVisionA11y';
+import {useCallbackRef} from './useCallbackRef';
+
+export function useWatchActiveSeries(
+  id: string,
+  onIndexChange: (event: CustomEventReturn) => void,
+) {
+  const onIndexChangeCallback = useCallbackRef(onIndexChange);
+
+  useEffect(() => {
+    const eventName = getEventName(id, COLOR_VISION_SINGLE_ITEM);
+
+    window.addEventListener(eventName, onIndexChangeCallback);
+
+    return () => {
+      window.removeEventListener(eventName, onIndexChangeCallback);
+    };
+  }, [id, onIndexChangeCallback]);
+}
--- a/packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
+++ b/packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
@@ -1,0 +1,107 @@
+import React from 'react';
+import type {Story} from '@storybook/react';
+import type {DataSeries} from '@shopify/polaris-viz-core';
+
+import {LineChart} from '../components/LineChart';
+import {PolarisVizProvider} from '../components/PolarisVizProvider';
+import {setSingleSeriesActive} from '../utilities';
+import {useWatchActiveSeries} from '../hooks';
+
+export default {
+  title: 'polaris-viz/Playground/ExternalEvents',
+  parameters: {},
+  decorators: [(Story) => <div>{Story()}</div>],
+  argTypes: {},
+};
+
+const DEFAULT_DATA: DataSeries[] = [
+  {
+    name: 'Breakfast',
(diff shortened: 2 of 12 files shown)


envex:
@nickpresta @philschoefer Take a look at this, the story can be found here: https://6062ad4a2d14cd0021539c1b-tsnlqfbano.chromatic.com/?path=/story/polaris-viz-playground-externalevents--default

packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+];
+
+const CHART_ID = 'chart_01';
+
+const Template: Story = () => {
+  const [activeName, setActiveName] = React.useState('none');
+
+  useWatchActiveSeries(CHART_ID, ({detail: {index, data}}) => {
+    if (index === -1 || data == null) {
```
nickpresta, line 62: When could `data` be `null` and when could `index` be `-1`?

If it's when the series is not active (e.g. moused outside the chart) I would almost expect the entire event to be `null` or the `detail` property to be `null`.

packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+    ],
+  },
+];
+
+const CHART_ID = 'chart_01';
+
+const Template: Story = () => {
+  const [activeName, setActiveName] = React.useState('none');
+
+  useWatchActiveSeries(CHART_ID, ({detail: {index, data}}) => {
```
nickpresta, line 61: There's an interesting bug that we're not understanding. When I hover over the button in Storybook, the event comes through with `detail: { index: 2 }` without a `data` key at all (not `undefined` or `null`).

I think we're not passing along `data` someplace we should be, which means even when you hover over the button, the text always says `Active series: none`

nickpresta (commented):
@philschoefer and I looked at this together. It's looking good so far, but we have a few questions.

packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -94,7 +94,7 @@ export function Chart({
   xAxisOptions,
   yAxisOptions,
 }: ChartProps) {
-  useColorVisionEvents(data.length > 1);
+  useColorVisionEvents(data.length > 1, data);
```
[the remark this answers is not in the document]
envex: The first parameter here enables/disables the events. It can be any boolean, not just data related.

packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
envex: I'm hesitant to pass the data back to consumers when triggering events outside of the chart. Ideally, they'd have the data they passed in anyways, so us passing the data prop back in events triggered from inside the chart is just a nice-to-have in case, for some reason, the data gets out of sync.

packages/polaris-viz/src/hooks/useWatchActiveSeries.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export function useWatchActiveSeries(
+  id: string,
+  onIndexChange: (event: EventReturn) => void,
+) {
+  const onIndexChangeCallback = useCallbackRef(onIndexChange);
+
+  useEffect(() => {
+    if (!id) {
```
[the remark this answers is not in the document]
envex: I've removed the check. It was copied over from the original function and typescript should catch it.

packages/polaris-viz/src/playground/ExternalEvents.stories.tsx
envex: @nickpresta What do you think about just removing `data` from the return until we find a need to get data back from the chart instead of using the data a consumer _should_ already have available?

nickpresta: I think that's a fine assumption for now. We already assume (and require) the consumer to look at their own `data` object, anyways.

PR #1470, opened 2023-01-25
Pass theme to tooltip content hooks

envex:

We weren't passing the theme to the `renderLinearComparisonTooltip` method, which was causing the theme to always be `Default (dark mode)`.
 
### Before merging


pbojinov (approved):
🎩 and code looks good ✅

PR #1480, opened 2023-02-08
Add TrendIndicator

envex:

Initial exploration for a TrendIndicator component that can be used in `web`.

### Why an SVG?

There was an ask about adding these TrendIndicators in charts like Funnel and SimpleBarChart.

image

To allow the TrendIndicator to be an inline element in the chart, and be copyable/printable, it needs to be an svg element. 

Thoughts @pbojinov @LBWright @nickpresta @philschoefer? Is it worth supporting an SVG element instead of just plain ol' HTML?

image
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-tneqaezlby.chromatic.com/?path=/story/polaris-viz-subcomponents-trendindicator--default

https://6062ad4a2d14cd0021539c1b-tneqaezlby.chromatic.com/?path=/story/chromatic-components--trendindicator

### DIFF
--- a/packages/polaris-viz-core/src/constants.ts
+++ b/packages/polaris-viz-core/src/constants.ts
@@ -225,9 +225,9 @@ export const DEFAULT_THEME: Theme = {
     labelColor: variables.colorGray30,
     backgroundColor: variables.colorGray150,
     trendIndicator: {
-      positive: '#03AB92',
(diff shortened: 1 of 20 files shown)


LBWright:
I'm a really big fan of this. We were wanting this for an upcoming project and I think it fits within PV.

PR #1491, opened 2023-02-13
Remove `hi` from `target_languages`

Al-Campuzano:
## This PR:
Removes Hindi (hi) from the `target_languages` in the component for  this repo. 
That language is no longer being used for any components that do not have the "audience" of `buyer`. See more here: https://github.com/Shopify/intl-languages/issues/190

Translation Platform is not requesting translations for that locale anyway, so it is unnecessary and misleading to have that in the `target_languages` list.

### DIFF
--- a/translation.yml
+++ b/translation.yml
@@ -7,7 +7,6 @@ target_languages:
     es,
     fi,
     fr,
-    hi,
     it,
     ja,
     ko,


envex (approved):

Al-Campuzano:
I just realized that this repo is no longer onboarded onto Translation Platform :derp:
So this is unnecessary. Closing

PR #1497, opened 2023-03-06
Stop the tooltip from covering the crosshair

envex:

The `<LineChart />` tooltip would try and follow the cursor around when interacting with the chart. This allowed the tooltip to get into a state where it was covering the crosshair for the active state, which covers the data.

Now, we're going to make the tooltip never cover the crosshair. It will always try and render on either side of the crosshair that keeps the tooltip within the bounds of the chart.

**Before**

https://user-images.githubusercontent.com/149873/223225806-da4465fa-f095-49fd-b0f7-cf32c8474a95.mov

**After**

https://user-images.githubusercontent.com/149873/223225878-fe9d4823-375d-40e1-8cb8-516b7d0da4e1.mov

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/LineChart/utilities/getAlteredLineChartPosition.ts
+++ b/packages/polaris-viz/src/components/LineChart/utilities/getAlteredLineChartPosition.ts
@@ -1,0 +1,141 @@
+import type {BoundingRect, Dimensions} from '@shopify/polaris-viz-core';
+import {clamp} from '@shopify/polaris-viz-core';
+
+import type {TooltipPositionOffset} from '../../TooltipWrapper';
+import type {Margin} from '../../../types';
+import {TooltipHorizontalOffset} from '../../TooltipWrapper';
+
+// The space between the cursor and the tooltip
+export const TOOLTIP_MARGIN = 20;
+
+export interface AlteredPositionProps {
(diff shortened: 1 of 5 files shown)


packages/polaris-viz/src/components/LineChart/utilities/getAlteredLineChartPosition.ts
```diff
@@ -0,0 +1,141 @@
...
+  const newPosition = {...position};
+
+  let x = currentX;
+  let y = currentY;
+
+  //
+  // Y POSITIONING
+  //
+
+  if (props.isPerformanceImpacted) {
```
pbojinov, line 45: Just curious, when does this prop become true?

pbojinov (approved):
Tophat and code looks good to me. Non-blocking but if you feel up to adding some tests on those util functions it could be use useful as a reference to the implementation or potential to catch any edge cases.

packages/polaris-viz/src/components/LineChart/utilities/getAlteredLineChartPosition.ts
envex: There's logic based on the amount of data points that will be rendered.

LBWright (approved):

PR #1500, opened 2023-03-07
Tweak TrendIndicator designs based on feedback

envex:

Based on feedback from @mirualves and @rachelng we're adding 2 different font-sizes for the `<TrendIndicator />`.

https://6062ad4a2d14cd0021539c1b-rfsqunquzq.chromatic.com/?path=/story/chromatic-components--trendindicator

### Character Width Changes

We're using a `semibold` font weight for the `TrendIndicator` which meant that the measurements were all wrong because we were measuring at the normal font weight. I've added `font-weight` offsets along with the existing `font-size` offsets.

Also, we previously allowed any decimal places that were returned by the browser, but I don't think we need that much precision so now we're rounding to 2 decimal places. This also helps a bit with file size.

image

Figma: https://www.figma.com/file/OmROczsf9FLj0ASdXhEOVx/Analytics-identity?node-id=204-51884&t=CDeTtgEde3tLkp90-0


pbojinov (approved):
Looks good! Couple non-blocking comments.

Also do we need a small `null data` trend indicator? It was hard to tell if that exists by looking the storybook

image

envex:
@pbojinov There's a small `null data`. It goes 2 big, 2 small, 2 big, 2 small.

PR #1501, opened 2023-03-08
WIP: Render missing data area

envex:

We're exploring rendering a visual empty area when benchmarks data isn't available.

image


tylernoseworthy (approved):
🚀

LBWright:
Is there any reason we wouldn't want this to be optional?

envex:
@LBWright Maybe in the future, but right now it's pretty tied to the `RelatedAreas` component as well.

Issue #1519, opened 2023-04-10
[Bug] Spark Line Chart 0 Data Height and Gradient is Inconsistent


clairedeboer:
## Bug summary

<img width="966" alt="Screenshot 2023-04-10 at 3 11 56 PM" src="https://user-images.githubusercontent.com/43480700/230985809-4efc775f-7ad6-40d0-98c3-772db3564d88.png">

The charts with 0 data have a line height and extra gradient that is not consistent with the charts that have data. 

I think this needs to be fixed is Polaris Viz. We investigated the issue and it is happening in Polaris Viz as well as in web leading me to think that it needs to be fixed in Polaris Viz. https://shopify.slack.com/archives/CNB58FZ34/p1679407433716979

## Expected behavior

Charts with 0 data should be consistent with charts that do have data. 

## Actual behavior

The chart for Online store conversion rate has 0 data but the line doesn't align with the Average order value chart which starts at 0 and the gradient is different. 

## Steps to reproduce the problem

1. First step: Choose a small shop such as the default shop in Spin
1. Second step: Go to this url: https://admin.web.web-minimal-u0qa.claire-deboer.us.spin.dev/store/shop1/marketing?attributionModel=last_click_non_direct

## Reduced test case

The best way to get your bug fixed is to provide a reduced test case. This [CodeSandbox template](https://codesandbox.io/s/shopify-polaris-viz-react-typescript-playground-zhkdqx) is a great starting point.

## Specifications

- Polaris-Viz version number:
- Browser:
- Device:
- Operating System:

Or run `npx envinfo --system --binaries --browsers --npmPackages react,react-dom,@shopify/polaris-viz` to provide specifications on your environment including version numbers, browser, device, and operating system.

Paste the results here:

```bash
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 505.27 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.1 - /opt/dev/sh/nvm/versions/node/v16.19.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.19.3 - /opt/dev/sh/nvm/versions/node/v16.19.1/bin/npm
    Watchman: 2022.07.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Safari: 16.3
  npmPackages:
    react: ^18.2.0 => 18.2.0 
[...]

PR #1524, opened 2023-04-13
Render SparkLineChart 0 yScale at bottom of chart

envex:

We're making two changes here. 

The first is instead of rendering all zero values in the vertical center of the chart, we'll render it at the bottom of the chart so it matches how other charts render mixed values.

Second, if the chart renders a regular and comparison line with all zero values, we're only going to render the regular line so the two lines don't overlap.

Resolves https://github.com/Shopify/polaris-viz/issues/1519

| Before  | After  |
|---|---|
|image|image|

### Before merging


LevonCross:
Awesome! This update looks good from the UX perspective. We will need to loop Adam in and ensure this makes its way into the Analytics Identity & standards work.

packages/polaris-viz-core/src/hooks/useFilteredSparkLineData.ts
```diff
@@ -0,0 +1,30 @@
...
+    return data.reduce<number[]>((prev, cur) => {
+      const numbers = cur.data
+        .map(({value}) => value)
+        .filter((value) => value !== null) as number[];
+      return prev.concat(...numbers);
+    }, []);
+  }, [data]);
+
+  const areAllZeros = useMemo(() => {
+    return !allNumbers.some((num) => num !== null && num !== 0);
```
michaelnesen, line 19: super nit: I feel like using `every` here would be easier to follow
```suggestion
    return allNumbers.every((num) => num === null || num === 0)
```

michaelnesen (approved):
LGTM 📈

packages/polaris-viz-core/src/hooks/useFilteredSparkLineData.ts
michaelnesen: Ohhh interesting 👂 👍

ryanrphillips:
Legends, thanks for the fast turnaround!

PR #1527, opened 2023-04-17
Allow MissingDataArea to render anywhere on the chart

envex:

It was assumed that the missing data for benchmarks would always be at the end of the chart, but that was wrong.

We want t be able to render the diagonal lines wherever the data is actually missing, instead of assuming the end.

### Data change needed for web

Currently in web [we are stripping `null` values from benchmarks](https://github.com/Shopify/web/blob/93d1c4f64612b5c02447bc2c95624c7a8e32d98b/app/sections/Analytics/ReportViewer/components/ReportViewer/components/Chart/components/LineChartRelational/LineChartRelational.tsx#L43), which originally resulted in arrays with different lengths. We don't want to do this anymore because it's also positioning the benchmark lines always at the beginning of the chart.

Resolves https://github.com/Shopify/core-issues/issues/54090
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-kqtjsvzndv.chromatic.com/?path=/story/polaris-viz-chromatic-charts-linechartrelational--missing-end-data

### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Allow `<MissingDataArea />` to render wherever values are `null` instead of at the end of the chart.
+
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


rita-morozova (approved):
🚀

PR #1531, opened 2023-04-24
Upgrade react-native-svg, change it to a peer dependency

fortmarek:

The primary reason I am making this PR is that `react-native-svg` is marked as a direct dependency of `@shopify/polaris-viz-native`. Since `react-native-svg` has native modules, we are effectively forcing the consumers the same version as we do since you can only resolve _one_ native module. To allow more flexibility, we should mark `react-native-svg` as a peer dependency – which should generally be the case for any dependency with native modules.

Whlie I was at it, I also updated `react-native-svg` to the latest version.

### Before merging


sandbox/package.json
```diff
@@ -4,30 +4,30 @@
...
-    "expo-font": "~10.0.4",
-    "expo-linking": "~3.0.0",
-    "expo-splash-screen": "~0.14.0",
-    "expo-status-bar": "~1.2.0",
-    "expo-web-browser": "~10.1.0",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
-    "react-native": "0.64.3",
-    "react-native-gesture-handler": "~2.1.0",
+    "expo": "^48.0.0",
```
envex, line 12: Can we split all these updates into a second PR, or do you think it's safe to do it all in one go?

fortmarek:
Feel free to continue this PR, I don't currently plan to get back to it.

PR #1539, opened 2023-06-13
Clamp negative widths to stop bars from rendering broken

envex:

When creating a new negative width to account for the left offset of labels, sometimes the bar would get a negative width.

This would cause the bar to rendering in a broken state.

| Before  | After  |
|---|---|
|image|image|
 
## Storybook link

- [Negative Values](https://6062ad4a2d14cd0021539c1b-kurslvbzpf.chromatic.com/?path=/story/polaris-viz-charts-simplebarchart-playground--skinny-trend)
- [Positive Values](https://6062ad4a2d14cd0021539c1b-ribxtinwbm.chromatic.com/?path=/story/polaris-viz-charts-simplebarchart-playground--positive-trend-overflow)

### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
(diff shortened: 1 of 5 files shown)


rita-morozova (approved):

michaelnesen (commented):
LGTM, seperate from this issue but could we also change the order of the `value` and `trend indicator` for negative values. 

Screenshot 2023-06-13 at 1 19 10 PM

For positive values the trend indicator is always shown after the value. Should it be the same for negative values as well?

PR #1554, opened 2023-07-05
Added ability to set default theme

envex:

The `Default` theme for polaris-viz is currently the dark theme. In web, we're setting almost all the components to `Light` now that we no longer use the dark theme.

With uplift, we've added another theme called `Uplift` that we can hide behind a flag.

Instead of replacing all the instances of light with a flag-flip, we're now able to set the default theme in the provider, which will trickle down to all the components in web.
 
### Before merging


pbojinov (approved):
✅

PR #1555, opened 2023-07-05
Add new prop to funnel chart for helper tooltip

juwanpetty:
This adds the ability to render a Tooltip component from web into the FunnelChart. We pass in an array of objects and use the key from the `data` object to decide which data point Tooltip will render on

image

[Spin URL](https://admin.web.web-funnel-chart-popover.juwan-petty.us.spin.dev/store/shop1/marketing/reports/tactic?since=2023-06-06&until=2023-07-05&attributionModel=last_click_non_direct&utm_campaign_name=Welcome%20-%20Quiz%3A%20Email%201%20%28WLNMcW%29&utm_campaign_medium=email&referring_channel=klaviyo) where the Tooltip is rendered

 
### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/FunnelChart/Chart.tsx
+++ b/packages/polaris-viz/src/components/FunnelChart/Chart.tsx
@@ -26,6 +26,7 @@ import {
 } from '../../constants';
 
 import {FunnelChartXAxisLabels, FunnelSegment} from './components/';
+import type {LabelHelpers} from './FunnelChart';
 
 const X_LABEL_OFFSET = 16;
 const NEGATIVE_LABEL_OFFSET = -4;
@@ -35,6 +36,7 @@ export interface ChartProps {
   xAxisOptions: Required<XAxisOptions>;
   yAxisOptions: Required<YAxisOptions>;
   dimensions?: Dimensions;
+  labelHelpers?: LabelHelpers[];
 }
 
 export function Chart({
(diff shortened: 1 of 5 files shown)


packages/polaris-viz/src/components/FunnelChart/Chart.tsx
```diff
@@ -209,6 +220,9 @@ export function Chart({
         width={width}
         height={drawableHeight}
         fill={`url(#${gradientId})`}
+        style={{
+          pointerEvents: 'none',
+        }}
```
juwanpetty, line 225: Previously the mask was preventing the Tooltip from being able to register a hover mouse event

packages/polaris-viz/src/components/FunnelChart/components/Label.scss
```diff
@@ -1,14 +1,17 @@
...
+  > span {
     padding: 0 4px;
     border-radius: '4px';
     border-radius: 4px;
   }
 
-  span:empty {
+  > span:empty {
     display: none;
   }
```
juwanpetty, line 16: Added more specificity with `>` because without it, it was also targeting the span element inside the Tooltip component

packages/polaris-viz/src/components/FunnelChart/components/Label.tsx
```diff
@@ -58,16 +61,18 @@ export function Label({
...
             color,
             lineHeight: `${HORIZONTAL_BAR_LABEL_HEIGHT}px`,
-            height: HORIZONTAL_BAR_LABEL_HEIGHT,
           }}
         >
           <span
             style={{
               backgroundColor,
+              height: HORIZONTAL_BAR_LABEL_HEIGHT,
+              userSelect: 'none',
```
juwanpetty, line 70: `userSelect: none` is to add back the same functionality we had where elements in the chart were non-selectable

PR #1562, opened 2023-07-14
Account for trend indicators with an undefined value when calculating SimpleBarChart horizontal scale

acmertz:

Trend indicators with an undefined `value` are not included in the horizontal scale calculation for `<SimpleBarChart>` and can overflow the edge of the chart.

Screenshot 2023-07-14 at 13 32 11

### Before

Screenshot 2023-07-14 at 15 03 30

### After

Screenshot 2023-07-14 at 15 02 11

## Storybook link

https://6062ad4a2d14cd0021539c1b-leqgvzknel.chromatic.com/?path=/docs/polaris-viz-charts-simplebarchart--with-trend-indicators

Inspect the chart container and confirm that the trend indicator is within the padded area:

Screenshot 2023-07-14 at 16 20 47

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/SimpleBarChart/utilities.ts
+++ b/packages/polaris-viz/src/components/SimpleBarChart/utilities.ts
@@ -21,10 +21,6 @@ export function getLongestTrendIndicator(
       for (const [index, trend] of trendEntries) {
         const dataPoint = seriesData[index];
 
-        if (trend.value == null) {
-          continue;
-        }
-
         if (dataPoint?.value === highestPositive) {
           longestTrendIndicator.positive = estimateTrendIndicatorWidth(
             trend.value,
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -11,6 +11,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Added support to hide overflowing LegendItems in vertical legends
 
+### Fixed
+
+- Fixed issue where trend indicators could extend past the chart bounds in `<SimpleBarChart />` when passing a value of `undefined` to `TrendIndicator.value`.
+
 ## [10.4.2] - 2024-02-12
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


packages/polaris-viz/src/components/SimpleBarChart/utilities.ts
```diff
@@ -21,10 +21,6 @@ export function getLongestTrendIndicator(
       for (const [index, trend] of trendEntries) {
         const dataPoint = seriesData[index];
 
-        if (trend.value == null) {
```
pbojinov, line 24: This was just added in https://github.com/Shopify/polaris-viz/pull/1560/files#diff-9ac33a141fa88f91f43ed900a667ec4876d0d408b5af8685ae07d6c646045f16R23-R27 to prevent an undefined value from crashing the SimpleBarChart

Not sure we want to remove it.

acmertz: @pbojinov good observation! This PR also updates `estimateTrendIndicator()` to support an undefined `value` parameter, in which case it uses the [NO_VALUE_WIDTH](https://github.com/Shopify/polaris-viz/blob/40c76ea06b1e5bfd3a51ccb6225a77fbd798ab8d/packages/polaris-viz/src/components/TrendIndicator/constants.ts#L4) constant as a fallback width (`NO_VALUE_WIDTH` is used [in TrendIndicator.tsx](https://github.com/Shopify/polaris-viz/blob/40c76ea06b1e5bfd3a51ccb6225a77fbd798ab8d/packages/polaris-viz/src/components/TrendIndicator/TrendIndicator.tsx#L44) as the width for trend indicators that don't have a value).

Data points with an undefined `value` appear to be working without issue in the Storybook for this PR: https://6062ad4a2d14cd0021539c1b-udbtypmgrc.chromatic.com/?path=/docs/polaris-viz-charts-simplebarchart--with-trend-indicators

acmertz: @pbojinov I added a new Chromatic story to make sure that this behavior is represented: e59b026

acmertz:
@pbojinov @envex This issue actually just came up again in https://github.com/Shopify/web/pull/118428#issuecomment-1948412028 - mind having another look? (sorry to resurrect such an old PR!)

envex (approved):

PR #1570, opened 2023-08-01
Allow TooltipWrapper to use a portal

envex:
For `<LineChart />` we're going to put the `<Tooltip />` in a portal so that it can be rendered outside the bounds of the chart.
 
## Storybook link

[BarChart Examples
](https://6062ad4a2d14cd0021539c1b-tzjouayhiu.chromatic.com/iframe.html?args=&id=polaris-viz-charts-barchart-playground--external-tooltip&viewMode=story)

[LineChart Examples](https://6062ad4a2d14cd0021539c1b-tzjouayhiu.chromatic.com/iframe.html?args=&id=polaris-viz-charts-linechart-playground--external-tooltip&viewMode=story)

Spin: https://admin.web.external-tooltips.matthew-vickers.us.spin.dev/store/shop1/dashboards

### Before merging

### DIFF
--- a/packages/polaris-viz-core/src/index.ts
+++ b/packages/polaris-viz-core/src/index.ts
@@ -150,4 +150,5 @@ export type {
   DataGroup,
   TargetLine,
   ErrorBoundaryResponse,
+  Position,
 } from './types';
--- a/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx
+++ b/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx
@@ -1,0 +1,73 @@
+import type {Story} from '@storybook/react';
+
+import type {BarChartProps} from '../../BarChart';
+import {BarChart} from '../../BarChart';
+import {META} from '../meta';
+
+export default {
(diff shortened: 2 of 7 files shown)


susiekims (approved):
Looks great, thanks!✨

rita-morozova (approved):

packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -234,12 +227,8 @@ export function Chart({
       });
 
       return {
-        x:
-          xScale(activeIndex) +
-          chartXPosition +
-          parseInt(selectedTheme.chartContainer.padding, 10),
-        y: svgY,
-        position: TOOLTIP_POSITION,
```
susiekims, line 242: I think we'll have to go back to using the active index position instead of the mouse position, I just noticed the tooltip position is wrong when i use tab to focus between the points:
image

envex: This case shouldn't be hit when using the keyboard. It will use the else on line 246.

I think we need to add the chart bounds like we do in the BarChart.

PR #1578, opened 2023-08-22
Fix stacked area arity bug

maryamkaka:

This PR fixes a bug we are seeing with the stacked area chart throwing when the length of the data changes. After some investigation we found that React spring is unable to animate svgs of different lengths without a custom interpolator but that seemed like overkill for now. Instead I've decided to just not animate when the length of data changes. 

I've updated the Dynamic Data story for the Stacked Area component to change data length as well as value.  

Part of the work for: https://github.com/Shopify/core-issues/issues/59450

🖼 Include screenshots of before and after, if relevant

Before: 

https://github.com/Shopify/polaris-viz/assets/13425309/342fe148-e3b3-442b-ba89-74c2c6a0d5ad

After:

https://github.com/Shopify/polaris-viz/assets/13425309/e06036cc-3239-4f2a-903b-97649de7d581

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/StackedAreaChart/components/Area/Area.tsx
+++ b/packages/polaris-viz/src/components/StackedAreaChart/components/Area/Area.tsx
@@ -1,86 +1,30 @@
-import {useRef} from 'react';
-import {animated, useSpring} from '@react-spring/web';
-import type {SpringValue} from '@react-spring/web';
-import type {Area as D3Area, Line} from 'd3-shape';
 import {
   LinearGradientWithStops,
   getColorVisionEventAttrs,
   COLOR_VISION_SINGLE_ITEM,
   getColorVisionStylesForActiveIndex,
-  useChartContext,
-  AREAS_LOAD_ANIMATION_CONFIG,
   getGradientFromColor,
-  useSpringConfig,
-  AREAS_TRANSITION_CONFIG,
 } from '@shopify/polaris-viz-core';
-import type {Color, Theme, GradientStop} from '@shopify/polaris-viz-core';
-
-import type {StackedSeries} from '../../../../types';
+import type {GradientStop} from '@shopify/polaris-viz-core';
 
 import styles from './Area.scss';
-
-export interface AreaProps {
-  activeLineIndex: number;
-  animationIndex: number;
-  areaGenerator: D3Area<number[]>;
-  colors: Color[];
-  data: StackedSeries;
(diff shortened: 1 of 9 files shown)


packages/polaris-viz/src/components/StackedAreaChart/components/StackedAreas/StackedAreas.tsx
```diff
@@ -84,8 +86,14 @@ export function StackedAreas({
   return (
     <Fragment>
       {stackedValues.map((data, index) => {
+        const dataIsValidForAnimation =
+          !previousStackedValues ||
+          data.length === previousStackedValues[index].length;
```
maryamkaka, line 91: This is a pattern I borrowed from `<LineSeries />` ([link to code])(https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz-core/src/components/LineSeries/LineSeries.tsx#L111-L114). I wasn't able to figure out a way to get react spring to not animate when the data length changes but I'm open to ideas if folks know of a better way that creating a whole new component 🙏

packages/polaris-viz/src/components/StackedAreaChart/components/Area/Area.tsx
```diff
@@ -1,84 +1,29 @@
...
-    to: {
-      opacity: 0.25,
-      animatedAreaShape: areaGenerator(data),
-      animatedLineShape: lineGenerator(data),
-    },
-    ...springConfig,
-  });
+  const opacity = 0.25;
+  const animatedAreaShape = areaGenerator(data);
+  const animatedLineShape = lineGenerator(data);
```
susiekims, line 26: Should these just be `areaShape` and `lineShape` since this is the non-animated `Area`?

packages/polaris-viz/src/components/StackedAreaChart/components/Area/Area.tsx
```diff
@@ -1,84 +1,29 @@
-import {useRef} from 'react';
-import {animated, useSpring} from '@react-spring/web';
-import type {SpringValue} from '@react-spring/web';
-import type {Area as D3Area, Line} from 'd3-shape';
+import {animated} from '@react-spring/web';
```
susiekims, line 1: Instead of using `animated.path`, could we just use `path`?

susiekims (approved):
Looks great!

PR #1587, opened 2023-09-19
Add tooltip support for Donut Chart

michaelnesen:

Adds tooltip support for `DonutChart`

Screenshot 2023-09-19 at 10 01 03 AM

 
## Storybook link
http://localhost:6006/?path=/story/polaris-viz-charts-donutchart--tooltip

### Before merging


packages/polaris-viz/src/components/DonutChart/DonutChart.tsx
```diff
@@ -65,6 +71,7 @@ export function DonutChart(props: DonutChartProps) {
         legendPosition={legendPosition}
         renderInnerValueContent={renderInnerValueContent}
         renderLegendContent={renderLegendContent}
+        renderTooltipContent={tooltipOptions ? renderTooltip : undefined}
```
michaelnesen, line 74: Currently, `tooltipOptions` is what enables the tooltip and defines its options in the donut chart. I believe we want the tooltip to be an opt-in experience cc @rachelng. But perhaps another prop like `showTooltip` would make more sense? Interested to see what others think

envex: I like the `showTooltip` prop. If we decide to go that route, we should add it to all the other components that can render tooltips.

michaelnesen:
> @michaelnesen Tooltips don't show when tabbing through the chart, is that intended?

Hmm good question, It looks like for all other charts tabbing through data points shows the tooltip. However with the donut chart it's a little different as tabbing through already selects/highlights the legend. I can check with UX to see if they have an opinion.  

However I'm currently not sure why the tooltip isn't showing up when the active index changes by focus 🤔 But if all other charts have this tooltip behaviour perhaps it makes sense to also have it in the donut chart for consistency

https://github.com/Shopify/polaris-viz/assets/90475364/18d4ffd6-9e39-4307-8939-e11f8fc787d8

packages/polaris-viz/src/components/DonutChart/Chart.tsx
```diff
@@ -208,60 +229,59 @@ export function Chart({
               viewBox={`${minX} ${minY} ${viewBoxDimensions.width} ${viewBoxDimensions.height}`}
               height={diameter}
               width={diameter}
+              ref={setSvgRef}
             >
-              {isLegendMounted && (
```
envex, line 212: I don't think the `isLegendMounted` check is used anymore and it was causing issues with attaching focus events to the arcs.

PR #1595, opened 2023-09-29
[Bug Fix]: Adjust tooltip position on StackedAreaChart when page is scrolled

euripidean:

Fixes: https://github.com/Shopify/core-issues/issues/60331

**Issue**
Received a possible bug that the `ReturningCustomerRate` card was not rendering a tooltip.

- Investigated and found that the tooltip _was_ being rendered, but that when the window was scrolled, the tooltip was rendering outside of the bounds of the chart and was no longer visible. It was also being partially cut off when the scroll value was lower.

https://github.com/Shopify/polaris-viz/assets/33559193/40426d92-6773-4952-b1ad-8458794af3df

- To replicate the issue, staff in to any store and navigate to the Dashboard. Ensure the `ReturningCustomerRate` card is displayed (currently the only StackedArea chart available). 
- If you move this card to a place on the dashboard and have to scroll to access it, hover over and you'll notice the tooltip does not appear in the expected position.
- Move the card to the top of the dashboard and hover over it. Without scroll, the tooltip will appear as expected.
- You can experiment with different scroll amounts to see the full behaviour.

Fixes https://github.com/Shopify/core-issues/issues/60331

Note: Tested in Storybook - need to confirm this fix works once updated and available in web.

https://github.com/Shopify/polaris-viz/assets/33559193/545208ca-22eb-4ba8-b462-9b4315b0fce7

## Extra Thoughts 
There are a few similar functions across the charts now to adjust the position of the tooltip in various scenarios. I was thinking it might be worth combining these into one function to adjust the positioning that (ideally) works for all of the charts. 

Happy to look into this further if there's consensus this needs to happen.

 
### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/StackedAreaChart/Chart.tsx
+++ b/packages/polaris-viz/src/components/StackedAreaChart/Chart.tsx
@@ -62,6 +62,7 @@ import {useStackedData} from './hooks';
 import {StackedAreas, Points} from './components';
 import {useStackedChartTooltipContent} from './hooks/useStackedChartTooltipContent';
 import {yAxisMinMax} from './utilities/yAxisMinMax';
+import {getAlteredStackedAreaChartPosition} from './utilities/getAlteredStackedAreaChartPosition';
 import styles from './Chart.scss';
 
 const TOOLTIP_POSITION: TooltipPositionOffset = {
@@ -382,6 +383,7 @@ export function Chart({
           focusElementDataType={DataType.Point}
           getMarkup={getTooltipMarkup}
           getPosition={getTooltipPosition}
+          getAlteredPosition={getAlteredStackedAreaChartPosition}
           id={tooltipId}
           margin={ChartMargin}
           onIndexChange={(index) => setActivePointIndex(index)}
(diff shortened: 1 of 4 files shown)


envex:
@euripidean You can setup a story in storybook to test the scrolling without having to put it in web: https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx

euripidean:
> @euripidean You can setup a story in storybook to test the scrolling without having to put it in web: https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx
> 
> We also have the [getAlteredLineChartPosition](https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz/src/components/LineChart/utilities/getAlteredLineChartPosition.ts) function you may be able to lean into.

Added the story. I did use the `getAlteredLineChartPosition` function as the basis for the `getAlteredStackedAreaChartPosition` function, but let me know if there's anything you think I missed.

pbojinov (approved):

PR #1600, opened 2023-11-06
Playing around with mismatched data

envex:
### Before merging


packages/polaris-viz/src/utilities/tests/fillMissingDataPoints.test.ts
```diff
@@ -0,0 +1,146 @@
...
+          {key: 'Birds', value: null},
+          {key: 'Lizards', value: null},
+          {key: 'Turtles', value: null},
+          {key: 'Snakes', value: 0},
+        ],
+      },
+    ]);
+  });
+
+  it('loops through a large data set in less than 10ms', () => {
```
envex, line 107: I'm not sure if this is even worth testing, but the original implementation below would get between `80` and `100ms`. The new implementation gets around `4ms`. 

```
for (const {data} of dataSeries) {
  for (const {key} of data) {
    allKeys.add(`${key}`);
  }
}

return dataSeries.map(({name, data}) => ({
  name,
  data: [...allKeys].map((key) => {
    const matchedValue = data.find((item) => item.key === key);
    return matchedValue ? matchedValue : {key, value: null};
  }),
}));
```

packages/polaris-viz/src/utilities/fillMissingDataPoints.ts
```diff
@@ -0,0 +1,36 @@
+import type {DataSeries} from '@shopify/polaris-viz-core';
+
+export function fillMissingDataPoints(dataSeries: DataSeries[]) {
+  const areAnyDataLengthsDifferent = dataSeries.some(
+    ({data}) => data.length !== dataSeries[0].data.length,
+  );
```
maryamkaka, line 6: Relying on this check to see if we need to fill in missing points is going to miss cases where the `dataSeries` is the same length but the `keys` are different. I think we'll want to check that the `key`s are the same across each series instead

e.g. This is a valid case where we'd want to continue but would early return 
``` [
  {
    name: 'Canada',
    data: [
      {key: 'Cats', value: 6.64},
      {key: 'Birds', value: 54.47},
    ],
  },
  {
    name: 'United States',
    data: [
      {key: 'Lizards', value: 350.13},
      {key: 'Turtles', value: 223.43},
    ],
  },
];
```

envex: Right, I thought about that but I'm not sure how often we'd hit this case.

I was originally doing an equality check on the next array in this `.some()` but it adds about 10ms onto the time.

I wonder if we should just always run the code and not try and bail early.

bencmilton: > I wonder if we should just always run the code and not try and bail early.

Agreed, probably worth just running it every time. Looking for extraneous keys would be a hard one to check too

```
  {
    name: 'Canada',
    data: [
      {key: 'Cats', value: 6.64},
      {key: 'Birds', value: 54.47},
    ],
  },
  {
    name: 'United States',
    data: [
      {key: 'Cats', value: 350.13},
      {key: 'Birds', value: 223.43},
      {key: 'Snakes', value: 100.00},
    ],
  },
];
```

PR #1612, opened 2023-12-05
[Donut Chart] Add legend with values + trend indicator data

michaelnesen:

This PR adds support for showing values and trends in `<DonutChart />` legend with new prop `showLegendValues`

https://github.com/Shopify/core-issues/issues/63578

Screenshot 2023-12-05 at 3 53 50 PM

 
### Before merging


packages/polaris-viz/src/components/DonutChart/components/LegendValues/LegendValues.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export function LegendValues({
+  data,
+  labelFormatter,
+  getColorVisionStyles,
+  getColorVisionEventAttrs,
+}: LegendContentProps) {
+  const selectedTheme = useTheme();
+
+  const maxTrendIndicatorWidth = data.reduce((maxWidth, {metadata}) => {
```
michaelnesen, line 26: Used this util `getTrendIndicatorData` to calculate the max trend indicator width so I could set the width of that column to the max trend with. Without this the trend indicators kept getting squished and the sizes were not consistent

Screenshot 2023-12-06 at 10 20 01 AM

michaelnesen: @envex Sorry, that photo was meant to show what it looked like without setting the width on the table column

michaelnesen: ah do you mean the spacing between the values and trends?

packages/polaris-viz/src/components/DonutChart/components/LegendValues/LegendValues.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const {trendIndicatorWidth} = getTrendIndicatorData(metadata.trend);
+
+    return Math.max(maxWidth, trendIndicatorWidth);
+  }, 0);
+
+  return (
+    <table className={styles.Table}>
+      {data.map(({name, data, metadata}, index) => {
+        const value = data[0].value || 0;
```
maryamkaka, line 39: Is it valid for us to default the absence of data as 0 🤔 

I think its probably fine in the case of the donut chart but wanted to throw it out there that I don't think that that assumption is always valid. I wonder if we might be better off rendering something like `-` or omitting it completely if we don't have information about it?

maryamkaka (commented):
Looking good! Just got a couple of small qs 👇

packages/polaris-viz/src/components/DonutChart/components/LegendValues/LegendValues.tsx
```diff
@@ -0,0 +1,95 @@
...
+            key={name}
+            style={{
+              ...getColorVisionStyles(index),
+            }}
+            {...getColorVisionEventAttrs(index)}
+          >
+            <td
+              style={{
+                width: '12px',
+              }}
```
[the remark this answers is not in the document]
michaelnesen: Each table data tag needed different styles (mainly width or padding) and weren't reusable so I just styled them inline. However I'm definitely not apposed to moving these into descriptive css classes, will make it more readable for sure 👍

packages/polaris-viz/src/components/DonutChart/components/LegendValues/LegendValues.tsx
michaelnesen: Yeah thats an interesting callout. Looking at other charts in Polaris viz, it seems when a data point has a value of null, no value is shown.

Screenshot 2023-12-11 at 8 58 12 PM

I think we should follow this pattern 👍 I'll check in with UX about if we want to omit the entire data point from the legend / maybe we just show nothing in the value cell.

packages/polaris-viz/src/components/DonutChart/Chart.tsx
```diff
@@ -94,7 +97,7 @@ export function Chart({
 
   const longestLegendWidth = data.reduce((previous, current) => {
     const estimatedLegendWidth = estimateLegendItemWidth(
-      current.name ?? '',
+      `${current.name ?? ''} ${current.data[0].value} `,
```
maryamkaka, line 100: Wouldn't we only want to add the data value if we are showing the value in the legend 🤔 Should we add a check to see if `showLegendValues == true`?

maryamkaka (approved):
🍩

PR #1620, opened 2024-01-18
[UA][Polaris-Viz]: Fixed Tooltip Positioning When There Are Many Bars

precillieo:
### Before merging


envex (changes requested):
Can we pull the tooltip above the chart so we're not covering the data?

envex (approved):

PR #1633, opened 2024-02-28
Allow id to be null

envex:

Allows a consumer to conditionally watch for events. If for some reason the `id` prop is not available, we won't attach the events to window.

### DIFF
--- a/packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
+++ b/packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
@@ -1,0 +1,94 @@
+import {mount} from '@shopify/react-testing';
+import {COLOR_VISION_SINGLE_ITEM} from '@shopify/polaris-viz-core';
+
+import {
+  useWatchActiveSeries,
+  setActiveSeriesListener,
+} from '../useWatchActiveSeries';
+import {getEventName} from '../ColorVisionA11y';
+import {useCallbackRef} from '../useCallbackRef';
+
+jest.mock('../useCallbackRef');
+
+const mockUseCallbackRef = useCallbackRef as jest.MockedFunction<
+  typeof useCallbackRef
+>;
+
+const onIndexChangeCallback = jest.fn();
+
+describe('useWatchActiveSeries()', () => {
+  beforeEach(() => {
+    mockUseCallbackRef.mockReturnValue(onIndexChangeCallback);
+  });
+
+  afterEach(() => {
+    mockUseCallbackRef.mockReset();
+  });
+
+  describe('useWatchActiveSeries()', () => {
+    it('attaches event listener to window when id is passed', () => {
+      const addEventListenerSpy = jest.spyOn(window, 'addEventListener');
+
+      function MockComponent() {
+        useWatchActiveSeries('id', onIndexChangeCallback);
+
+        return null;
+      }
+
+      mount(<MockComponent />);
+
+      expect(addEventListenerSpy).toHaveBeenCalledWith(
+        'id:color-vision-event:singleItem',
+        onIndexChangeCallback,
+      );
+    });
+
+    it('removes event listener from window when component is unmounted', () => {
+      const onIndexChangeCallback = jest.fn();
+      const removeEventListenerSpy = jest.spyOn(window, 'removeEventListener');
+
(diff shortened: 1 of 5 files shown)


kvendrik (changes requested):
If ID is `null` it won't do anything which makes it weird that you'd call it in the first place. How about we create a second export that makes all of this available as a non-hook? Based on [the logic](https://github.com/Shopify/polaris-viz/blob/bcf9e6a381cc4046047eee03e9477253787c434e/packages/polaris-viz/src/hooks/useWatchActiveSeries.ts) that seems to be possible and easy enough to do?

Alternatively you could export `getEventName()` and let consumers implement custom logic that does the same thing but outside of a hook

packages/polaris-viz/src/hooks/useWatchActiveSeries.ts
```diff
@@ -7,18 +7,34 @@ import {getEventName} from './ColorVisionA11y';
 import {useCallbackRef} from './useCallbackRef';
 
 export function useWatchActiveSeries(
-  id: string,
+  id: string | null,
```
kvendrik, line 10: The point of adding the separate `setActiveSeriesListener()` is so we don't need to make this optional right? So let's remove that logic?

packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
```diff
@@ -0,0 +1,44 @@
+import {mount} from '@shopify/react-testing';
+
+import {useWatchActiveSeries} from '../useWatchActiveSeries';
+
+describe('useWatchActiveSeries', () => {
```
kvendrik, line 5: - Nit: `useWatchActiveSeries()` as the describe as it's a listener
- Should have a nested `useWatchActiveSeries()` and `setActiveSeriesListener()` so that we test both

packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const addEventListenerSpy = jest.fn((_) => {});
```
kvendrik, line 6: Just `jest.fn()` will do the same thing as what you have here

packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  beforeEach(() => {
+    jest
+      .spyOn(window, 'addEventListener')
+      .mockImplementation(addEventListenerSpy);
+  });
```
kvendrik, line 12: Define these in each test so it's clear what needs to go in to get something specific out

packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
kvendrik:
```tsx
describe('useWatchActiveSeries()', () => {
  describe('setActiveSeriesListener()', () => {
    it('generates an event name using the given ID', () => {
      const id = getEventName('someChartId', COLOR_VISION_SINGLE_ITEM);
      const addEventListenerSpy = jest.spyOn(window, 'addEventListener');
      setActiveSeriesListener('someChartId', () => {});
      expect(addEventListenerSpy).toHaveBeenCalledWith(id, expect.any(Function));
    });
  });
  describe.todo('useWatchActiveSeries()');
});
```

kvendrik (changes requested):

packages/polaris-viz/src/hooks/tests/useWatchActiveSeries.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+      setActiveSeriesListener('someChartId', () => {});
+
+      expect(addEventListenerSpy).toHaveBeenCalledWith(
+        id,
+        expect.any(Function),
+      );
+    });
+  });
+});
```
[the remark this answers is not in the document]
envex: @kvendrik Added a test for `removeEventListener`, let me know if you think it makes sense to add one to ensure `setActiveSeriesListener` returns a function as well.

kvendrik (approved):
Thank you! 🎉

PR #1640, opened 2024-03-19
fix: Expose Uplift theme

maael:

We need to be able to have two variations of the Uplift theme: a standard and a "ghost" version.

As such we can't simply overwrite properties of the theme as shown [here](https://polaris-viz.shopify.com/iframe.html?viewMode=docs&id=shared-themes-customizing--page#titleAnchor-81).

Instead we need to define multiple themes as shown [here](https://polaris-viz.shopify.com/iframe.html?viewMode=docs&id=shared-themes-customizing--page#titleAnchor-86).

To do this with Uplift as a base however, the theme needs to be exposed, which it currently isn't.

In theory this can be retrieved via the `@shopify/polaris-viz-core`, but the same is true for the other themes and that would mean that dependency becomes a direct dependency of the project, rather than through `@shopify/polaris-viz`.

Instead, it's proposed here that we just expose the Uplift theme in a manner similar to the others.

Related:
- https://github.com/Shopify/order-risk/issues/1694

There are no visual changes, this is a package API change.
 
## Storybook link

There are no visual changes, this is a package API change.

### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


envex (approved):

envex:
@maael `@shopify/polaris-viz@11.1.0` should have your changes

PR #1668, opened 2024-05-27
Use masks for areas with data instead of finding missing data

envex:

We were running into an issue where we would get a single data point with a value in between chunks of missing data. This would throw off the missing data checks because we didn't account for single data points.

Instead of trying to figure out what data was missing (but only when consecutive null values were present), which was getting confusing, we can just mask out the areas that have actual data.

We're also only checking for missing data for the `relatedIndex` data series instead of looping over all the data sets unnecessarily.

Resolves https://github.com/Shopify/core-issues/issues/71169

### Before merging


blboyle (approved):
Looks good to me!

PR #1684, opened 2024-06-26
HACK: We be formatting

envex:
https://6062ad4a2d14cd0021539c1b-hrdaqlealu.chromatic.com/?path=/story/polaris-viz-charts-barchart--default

### DIFF
--- a/packages/polaris-viz/src/components/BarChart/stories/Default.stories.tsx
+++ b/packages/polaris-viz/src/components/BarChart/stories/Default.stories.tsx
@@ -8,8 +8,63 @@ import {DEFAULT_DATA, Template} from './data';
 
 export const Default: Story<BarChartProps> = Template.bind({});
 
+const data = [
+  {
+    name: 'Monday',
+    data: [
+      {key: 'Breakfast', value: 3},
+      {key: 'Lunch', value: 4},
+      {key: 'Dinner', value: 7},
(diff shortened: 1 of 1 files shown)


packages/polaris-viz/src/components/BarChart/stories/Default.stories.tsx
```diff
@@ -8,8 +8,63 @@ import {DEFAULT_DATA, Template} from './data';
...
+  {
+    name: 'Thursday',
+    data: [
+      {key: 'Breakfast', value: 33},
+      {key: 'Lunch', value: 34},
+    ],
+  },
+];
+
+function formatData(data) {
```
envex, line 44: Format the data back to what polaris-viz expects.

packages/polaris-viz/src/components/BarChart/stories/Default.stories.tsx
```diff
@@ -8,8 +8,63 @@ import {DEFAULT_DATA, Template} from './data';
 
 export const Default: Story<BarChartProps> = Template.bind({});
 
+const data = [
+  {
+    name: 'Monday',
```
envex, line 13: Easier to reason about data structure where the top level `name` would be the primary dimension/group and then the `series.data` array are the bars.

PR #1693, opened 2024-07-16
Legend overflow fixes

susiekims:

* Fixes bug where legend items were hidden even thought there's enough room to render more
* Styling updates: removes background from legend items + centers legend container

The legend overflow calculation was using the `MINIMUM_LEGEND_ITEM_WITH_VALUE_WIDTH` instead of `MINIMUM_LEGEND_ITEM_WIDTH` when values were hidden. But I also just learned that [horizontal legends with values look like this](https://github.com/Shopify/polaris-viz/pull/1167), so we don't even need `MINIMUM_LEGEND_ITEM_WITH_VALUE_WIDTH`.
image

Resolve https://github.com/Shopify/core-issues/issues/73670
One of feedback items we received from CSS here: https://github.com/Shopify/core-issues/issues/72480

Legend items should shorten to `100px` before items start being hidden

https://github.com/user-attachments/assets/12c569a3-c3ea-422a-933f-6efb242cfce4

 
## Storybook link

https://6062ad4a2d14cd0021539c1b-abxwzkkvsr.chromatic.com/ 

### Before merging

### DIFF
--- a/packages/polaris-viz-core/src/constants.ts
+++ b/packages/polaris-viz-core/src/constants.ts
@@ -373,7 +373,7 @@ export const LIGHT_THEME: Theme = {
   legend: {
     valueColor: variables.colorGray160,
     labelColor: variables.colorGray100,
-    backgroundColor: variables.colorGray10,
+    backgroundColor: null,
     trendIndicator: {
       positive: TREND_POSITIVE,
       negative: TREND_NEGATIVE,
--- a/packages/polaris-viz/src/components/Legend/Legend.tsx
+++ b/packages/polaris-viz/src/components/Legend/Legend.tsx
@@ -1,4 +1,4 @@
-import {Fragment} from 'react';
+import {Fragment, useCallback} from 'react';
 import type {RefObject} from 'react';
 import type {LabelFormatter} from '@shopify/polaris-viz-core';
 import {DEFAULT_THEME_NAME} from '@shopify/polaris-viz-core';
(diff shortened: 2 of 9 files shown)


packages/polaris-viz/src/components/Legend/Legend.tsx
```diff
@@ -36,6 +36,15 @@ export function Legend({
...
 
+  const onDimensionChange = useCallback(
+    (index, dimensions: LegendItemDimension) => {
+      if (itemDimensions?.current) {
+        itemDimensions.current[index] = dimensions;
+      }
+    },
+    [itemDimensions],
+  );
+
```
susiekims, line 47: This isn't related to the bug, I just noticed that there would be a `Maximum stack depth has been reached` React error occasionally in LegendItem since `onDimensionChange` changed on every render 😅

packages/polaris-viz-core/src/constants.ts
```diff
@@ -373,7 +373,7 @@ export const LIGHT_THEME: Theme = {
   legend: {
     valueColor: variables.colorGray160,
     labelColor: variables.colorGray100,
-    backgroundColor: variables.colorGray10,
+    backgroundColor: variables.colorGray00,
```
envex, line 376: Should we allow this to be `null` so it doesn't render any color instead of having to match the background of the chart?

envex (approved):

packages/polaris-viz-core/src/constants.ts
susiekims: Good idea! Done ✅

michaelnesen (approved):
LGTM 🚀

envex:
@susiekims Don't forget to update the `CHANGELOG`

PR #1704, opened 2024-07-22
Legend styling changes

susiekims:

Follow up of https://github.com/Shopify/polaris-viz

- Change font size for legends, labels, and tooltips to 11px
- Remove focus state onClick in `LegendItem` (should still be visible for keyboard interaction)

| Before  | After  |
|---|---|
| image  | image |

 
## Storybook link

https://6062ad4a2d14cd0021539c1b-yrsertjuva.chromatic.com/?path=/story/intro--page

* Confirm the body text in legends, tooltips, annotations, axis labels, etc are 11px
* Confirm you do not see the blue outline focus state when clicking on a legend item, but you can see it when focusing on it via keyboard

### Before merging

### DIFF
--- a/packages/polaris-viz-core/CHANGELOG.md
+++ b/packages/polaris-viz-core/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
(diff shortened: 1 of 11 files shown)


brendo:
With this change, does the character widths file need to be regenerated ([docs](https://github.com/Shopify/polaris-viz/tree/main/scripts/character-widths)).

envex:
@brendo It shouldn't need to be updated if only the font-size is changed.

There's the character-widths file and also a character-offsets file that handles the change between font-sizes.

The character-width file really only needs to be updated now if the font-family changes.

brendo:
Ah great! Thanks for clarifying :)

PR #1706, opened 2024-07-25
add useTrendINdicators prop for simple normalized bar chart

rita-morozova:
### Before merging


packages/polaris-viz/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Added `useTrendIndicators` prop to `SimpleNormalizedBarChart` to support rendering the `TrendIndicator` component. This allows for a more flexible representation of data trends, accommodating designs that require more detailed trend indicators.
```
rita-morozova, line 12: I'm not sure if this is the right approach, but we cannot replace the chevrons with the trend indicator as it would disrupt the existing reports' experience. Also, I don't think we can replace the breakdown bar viz to use a stacked horizontal bar instead of a simple normalized bar 🤷‍♀️ 
I'm open to other suggestions.

cc @envex @tylernoseworthy

packages/polaris-viz/src/components/ComparisonMetric/ComparisonMetric.tsx
```diff
@@ -16,6 +18,18 @@ export function ComparisonMetric({
   accessibilityLabel,
 }: Omit<ComparisonMetricProps, 'dataIndex'>) {
   const selectedTheme = useTheme();
+  const useTrendIndicator = selectedTheme.trendIndicator.useTrendIndicator;
+
+  if (useTrendIndicator) {
```
envex, line 23: Would you mind checking with UX, we may be able to just use this everywhere and drop the chevron

packages/polaris-viz/src/components/ComparisonMetric/ComparisonMetric.tsx
```diff
@@ -15,41 +14,26 @@ export function ComparisonMetric({
...
-            fill={selectedTheme.legend.trendIndicator.negative}
-          />
-          <span style={{color: selectedTheme.legend.trendIndicator.negative}}>
-            {metric}
-          </span>
-        </span>
-      );
+      direction = 'downward';
+      break;
+    case 'neutral':
```
rita-morozova, line 25: @envex just wanted to confirm we don't want to display any arrows for the "neutral" trend. We want to display `–`?

envex: What about displaying nothing. Here for example, maybe it makes sense to not render the `-` at all

image

envex (approved):
Just the one comment, the rest is 👍

packages/polaris-viz/src/components/ComparisonMetric/ComparisonMetric.tsx
rachelng: I think we should keep it as is for now but worth thinking about in the future. Removing it and displaying nothing will create odd spacing/padding when other values have a negative/positive trend.

PR #1709, opened 2024-07-30
Render single data point values for comparison line chart

rita-morozova:

We have cases in the comparison data array where one non-null value is followed by a null value; this point is not rendered on the screen because the size of the dash stroke is too small. In order for the comparison line to be visualized on the screen, we need to have at least two consecutive non-null values or set the stroke to "none." My approach to fix it is to render these values as individual standing points instead of as a part of the line.

No

Before:

https://github.com/user-attachments/assets/0593cad8-5db5-431b-8679-6ab86ace67a3

After:

https://github.com/user-attachments/assets/b83735b4-7887-4505-a67a-4e33d0c7843d

## Storybook link

[Storybook](https://6062ad4a2d14cd0021539c1b-zzykkiadxr.chromatic.com/?path=/story/polaris-viz-charts-linechart--with-null-values)

- Make sure the standalone values in the comparison chart are rendered as the individual points.

### Before merging

### DIFF
--- a/packages/polaris-viz-core/src/components/LineSeries/LineSeries.tsx
+++ b/packages/polaris-viz-core/src/components/LineSeries/LineSeries.tsx
@@ -137,6 +137,34 @@ export function LineSeries({
 
   const zeroLineY = yScale(0);
 
+  /* Renders visual points for isolated data points in a comparison dataset.
+  The dash stroke is too small to render these points effectively. */
+  const renderIsolatedPoints = (
+    data: LineChartDataSeriesWithDefaults,
+  ): JSX.Element[] => {
+    const points: JSX.Element[] = [];
+    data.data.forEach((point, index) => {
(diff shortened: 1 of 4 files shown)


rita-morozova:
@rachelng could you give approval or disapproval on this approach? We want to render standalone values for comparisons as separate points, otherwise, the dash stroke is too thin to be displayed on the screen as a single point, and we end up having a comparison point only on hover like in this [issue](https://github.com/Shopify/core-issues/issues/71927)

rachelng:
Tagging @LevonCross to this as he's currently working on data states including null data.

blboyle (approved):
Looks good!

rita-morozova:
@envex I refactored this PR, and now it is not related to any currently open issue (I misread the ticket 🤣), but it fixes a current bug we have (see description). I confirmed with @rachelng  that we should fix it

PR #1723, opened 2024-09-13
Format annotation keys

envex:
Format `<Annotation />` keys with the xAxis formatter so that the keys provided and keys displayed are the same and don't rely on the raw data.


michaelnesen (approved):
Thanks for the quick turnaround! 🎩 LGTM

PR #1726, opened 2024-09-19
Fix crosshair position for a single data point in line chart

rita-morozova:

When rendering a line chart with single data points per series, the crosshair was not aligned with the actual data points. This misalignment was caused by the animation logic. I am adding a condition to bypass the animation logic if all series have only one data point.  Animating a single point or very sparse data does not provide any additional visual benefit to the user. 

Related https://github.com/Shopify/core-issues/issues/71927

| Before | After |
|--------|------|
|Screenshot 2024-09-18 at 2 46 26 PM|Screenshot 2024-09-18 at 2 46 54 PM|
|Screenshot 2024-09-18 at 2 46 05 PM|Screenshot 2024-09-18 at 2 47 35 PM|

 
## Storybook link

[Storybook](https://6062ad4a2d14cd0021539c1b-kpdjwutsmi.chromatic.com/?path=/story/polaris-viz-charts-linechart-playground--crosshair-offset)

- Make sure the crosshair is positioned correctly
- Test other line charts and make sure nothing is broken

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/LineChart/components/PointsAndCrosshair/PointsAndCrosshair.tsx
+++ b/packages/polaris-viz/src/components/LineChart/components/PointsAndCrosshair/PointsAndCrosshair.tsx
@@ -70,6 +70,13 @@ export function PointsAndCrosshair({
     }
     const offset = isCrosshair ? selectedTheme.crossHair.width / 2 : 0;
 
+    const allSingleDataPoints = data.some((series) => series.data.length !== 1);
+
+    // Bypass animation if there's only one data point
+    if (!allSingleDataPoints) {
+      return xScale(activeIndex == null ? 0 : activeIndex) - offset;
+    }
+
     if (
       animatedCoordinates != null &&
       animatedCoordinates[longestSeriesIndex] != null &&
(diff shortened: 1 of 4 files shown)


packages/polaris-viz/src/components/LineChart/components/PointsAndCrosshair/PointsAndCrosshair.tsx
```diff
@@ -70,6 +70,15 @@ export function PointsAndCrosshair({
     }
     const offset = isCrosshair ? selectedTheme.crossHair.width / 2 : 0;
 
+    const allSingleDataPoints = data.every(
```
envex, line 73: Do you think using `.some` instead so we're not always checking the entire data set would be better, or not worth the optimization?

```
const allSingleDataPoints = data.some(
  (series) => series.data.length !== 1,
);
```

rita-morozova: @envex now I'm a little confused by the logic of using `some` in this case. We need to check that every series in the data array has one value, like in this example:
```
const DATA= [
  {
    "name": "Apr 1 – Apr 14, 2020",
    "data": [
      {
        "value": 8,
        "key": "2020-04-02T12:00:00"
      },
    ]
  },
  {
    "name": "May 1 – May 14, 2020",
    "data": [
      {
        "value": 10,
        "key": "2020-04-02T12:00:00"
      },
    ]
  },
  {
    "name": "Jun 1 – Jun 14, 2020",
    "data": [
      {
        "value": 15,
        "key": "2020-04-02T12:00:00"
      },
    ]
  },
]
```
If only some of them have a single data point and some have several data points, we don't want to bypass the animation

PR #1729, opened 2024-09-25
Change logic around trend indicator width offset

envex:

When building the width offset values for trends, we were originally trying to associate a trend with the longest value in each series.

This broke down when a series contained a value that was longer than the series that contained the trend.

Now we're going to find the longest trend for each positive and negative side of the chart and use those values.

https://github.com/Shopify/core-issues/issues/77311

| Before  | After  |
|---|---|
|image|image|

 
### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


packages/polaris-viz/CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Changed logic around how we determine the longest trend width. We now measure the longest trend with for each positive and negative side without associating with the longest date series value.
```
rita-morozova, line 12: measure the longest trend  **width**

rita-morozova (approved):

PR #1730, opened 2024-09-25
Add startAxisLabel value to Annotations for formatted key matching

envex:

To support https://github.com/Shopify/web/pull/140219 we [switched to formatting the `Annotation.startKey`](https://github.com/Shopify/polaris-viz/pull/1723) values so when the `xAxisOptions.labelFormatter` would format the dates different, we could correctly target those labels.

We've had a few instances where consumers were relying on the unformatted values. The screenshot below for example, they were relying on using the index of `8` to render the annotation, while using the `xAxisOptions.labelFormatter` to render blank labels.

image

In order to allow both use cases, we're adding an `startAxisLabel` that will target the formatted labels and switching back to `startKey` being the unformatted value used in `DataSeries[]`.
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-qltxxnwjyc.chromatic.com/?path=/story/polaris-viz-charts-barchart-playground-unformattedannotationkeys--unformatted-annotation-keys

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/Annotations/Annotations.tsx
+++ b/packages/polaris-viz/src/components/Annotations/Annotations.tsx
@@ -52,26 +52,46 @@ export function Annotations({
 
     const annotations = Object.keys(annotationsLookupTable)
       .map((key) => {
-        const annotation = annotationsLookupTable[key];
+        const annotation: Annotation | null = annotationsLookupTable[key];
 
         if (annotation == null || annotation.axis === 'y') {
           return null;
         }
 
-        const formattedKey = labelFormatter(key);
-
-        if (!formattedLabels.includes(formattedKey)) {
+        if (annotation.startAxisLabel == null && annotation.startKey == null) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            'Annotation provided with no startAxisLabel or startKey property',
+            {annotation},
+          );
           return null;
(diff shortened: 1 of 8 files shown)


envex:
@philschoefer Yeah, that's the thought.

`startAxisLabel` is the formatted xAxis label and then we move back to `startKey` being the unformatted value passed in with `DataSeries`.

packages/polaris-viz/src/components/Annotations/Annotations.tsx
```diff
@@ -52,26 +52,46 @@ export function Annotations({
...
     const annotations = Object.keys(annotationsLookupTable)
       .map((key) => {
-        const annotation = annotationsLookupTable[key];
+        const annotation: Annotation | null = annotationsLookupTable[key];
 
         if (annotation == null || annotation.axis === 'y') {
           return null;
         }
 
-        const formattedKey = labelFormatter(key);
```
michaelnesen, line 61: Does this mean the `startAxisLabel` passed in will need to be formatted by the consumer? Would a boolean prop to toggle formatting of the key (as @philschoefer suggested) be easier? That way the consumer wouldn't have to worry about running this `startAxisLabel` through their label formatter before, they could just toggle it and polaris viz would take care of the formatting

PR #1735, opened 2024-10-10
Show error state when chart contains infinity

carysmills:
Fixes this broken chart we saw on Sidekick this morning, which results from us passing the chart a value of Infinity
Screenshot 2024-10-10 at 10 25 46 AM

Using the existing error handling in the chart, we can just make the chart show the error state. An error will be thrown, so we should see these errors in bugsnag if the consuming project surfaces them.

Note: I spent a while trying to add a test for this but had issues checking for the error. Please let me know if you have any ideas!

Resolves https://github.com/Shopify/web/issues/144056

| Before  | After  |
|---|---|
| Screenshot 2024-10-10 at 2 59 49 PM  | Screenshot 2024-10-10 at 2 59 36 PM  |
 
## Storybook link
You can check the storybook I added: http://localhost:6006/?path=/story/polaris-viz-charts-barchart-playground--infinity-data

### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## Unreleased
+
+- Fixed an issue where the bar chart collapsed if a value of Infinity was passed to it.
+
 ## [15.0.4] - 2024-09-26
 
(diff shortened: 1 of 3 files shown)


carysmills:
I see the test failure and will check that this doesn't break charts when there's empty data 👀

dahukish (approved):
looks good from my PoV 🙇

envex (approved):
👍 

Do you think it makes sense to do this check in all the charts? (Not in this PR)

philschoefer (approved):

Issue #1736, opened 2024-10-11
Ensure charts show error state when Infinity or NaN is passed as a value


carysmills:
## Bug summary
Followup to https://github.com/Shopify/polaris-viz/pull/1735

**Charts to fix:**
- [x] Vertical bar chart (does horizontal chart need to be checked separately?)
- [ ] Line chart: a bit broken but not as broken as bar chart
- [ ] Line chart relational: (broken)
- [ ] Line chart predictive: (broken)
- [ ] Horizontal bar chart (broken)
- [ ] Area chart (broken)
- [ ] Stackedbar chart

**Do not use useYScale:**
- [ ] SimpleNormalizedChart (renders blank, renders error if showing legends)
- [ ] Funnel chart (renders something broken)
- [ ] Combo chart (broken)
- [ ] SimpleBarChart (broken)
- [ ] SparkBarChart (renders but not the infinity bar)
- [ ] SparkLineChart (renders a line)

**Fixed:**
- [x] Donut chart (errors as expected)

## Expected behavior
Charts should show the existing error state if Infinity or NaN is passed as a value

## Actual behavior

The bar chart broke when Infinity was passed to it. See the linked PR for images. We need to assess the other charts.

PR #1739, opened 2024-10-17
Fix handling of Infinity data in all charts

envex:
We've discovered that sometimes our charts are being passed Infinity now and this fixes the visual issues. When Infinity is passed to the charts, we instead render the existing error state. We previously fixed this in the vertical bar chart - this PR fixes it across all the other charts.

It looks like a lot of code changes, but mostly it's added Stories to ensure the fixes work. The actual code changes are contained in the first commit. 

The code changes:
* when possible, take advantage of existing min/max values to perform checks, so we aren't adding performance bottlenecks
* the check in useYScale required us to reconsider our use of Infinity is as fallback value when we create scales in some charts, so you'll see that in a few places (including the changes to `clamp`)
* useYScale takes care of several charts but others need the checks internally
* some charts were not being passed the skeleton type so those are added

Followup to https://github.com/Shopify/web/issues/144056
Closes https://github.com/Shopify/polaris-viz/issues/1736

### Infinity states
| Before  | After  |
|---|---|
| chromatic-infinity-data--default  |  chromatic-infinity-data--default-NEW |

### Data series with empty array of data (providing this because it is related to the changes and is something we should watch carefully)
| Before  | After  |
|---|---|
| chromatic-empty-data-series--default |chromatic-empty-data-series--default-NEW |

### Data prop that is an empty array (providing this because it is related to the changes and is something we should watch carefully, although technically this doesn't satisfy types so maybe it's not something we should be _too_ concerned about)
| Before  | After  |
|---|---|
|chromatic-empty-data--default|chromatic-empty-data--default-new |
 
## Storybook link
You can check out added stories under the Chromatic section after running `yarn storybook:all`

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/FunnelChart/Chart.tsx
+++ b/packages/polaris-viz/src/components/FunnelChart/Chart.tsx
@@ -59,7 +59,7 @@ export function Chart({
 
   const maxValue = yValues.length === 0 ? 0 : Math.max(...yValues);
 
-  if (!isFinite(maxValue)) {
+  if (!Number.isFinite(maxValue)) {
     throw new Error('Data must be finite');
   }
 
--- a/packages/polaris-viz-core/src/hooks/useSparkBar.ts
+++ b/packages/polaris-viz-core/src/hooks/useSparkBar.ts
@@ -65,13 +65,13 @@ export function useSparkBar({
   const filteredData = removeNullValues(data[0]);
   const [defaultData] = data;
 
-  const hasInfiniteData = filteredData.some((value) => !isFinite(value));
+  const hasInfiniteData = filteredData.some((value) => !Number.isFinite(value));
 
   if (hasInfiniteData) {
     throw new Error('Data must be finite');
   }
 
-  if (!isFinite(targetValue)) {
+  if (!Number.isFinite(targetValue)) {
     throw new Error('Target value must be finite');
   }
 
--- a/packages/polaris-viz-core/src/hooks/useSparkLine.ts
+++ b/packages/polaris-viz-core/src/hooks/useSparkLine.ts
@@ -20,23 +20,23 @@ export function useSparkLine({
     data.map(({data}) => data.map(({value}) => value)),
   );
 
-  const minYValues = yValues.length ? Math.min(...yValues) : 0;
-  const maxYValues = yValues.length ? Math.max(...yValues) : 0;
+  const minYValue = yValues.length ? Math.min(...yValues) : 0;
+  const maxYValue = yValues.length ? Math.max(...yValues) : 0;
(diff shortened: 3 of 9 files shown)


packages/polaris-viz/src/components/FunnelChart/Chart.tsx
```diff
@@ -57,6 +57,10 @@ export function Chart({
   const xValues = dataSeries.map(({key}) => key) as string[];
   const yValues = dataSeries.map(({value}) => value) as [number, number];
 
+  if (!isFinite(Math.max(...yValues))) {
```
carysmills, line 60: We're doing another `Math.max(...yValues)])` further down for creating the yScale. Can we create a variable above for the check and the yScale so we're not doing this twice?

packages/polaris-viz/src/components/ChartContainer/components/ChartDimensions/ChartDimensions.tsx
```diff
@@ -55,7 +56,22 @@ export function ChartDimensions({
...
+  function handleError(error: Error, errorInfo: ErrorInfo) {
+    setHasError(true);
+
+    if (onError == null) {
+      onErrorProvider?.(error, errorInfo);
+      return;
+    }
+
+    onError(error, errorInfo);
+  }
```
carysmills, line 68: why do we need this now?

packages/polaris-viz/src/components/FunnelChart/Chart.tsx
```diff
@@ -55,7 +55,13 @@ export function Chart({
   const [maskRef, setMaskRef] = useState<SVGMaskElement | null>(null);
 
   const xValues = dataSeries.map(({key}) => key) as string[];
-  const yValues = dataSeries.map(({value}) => value) as [number, number];
+  const yValues = dataSeries.map(({value}) => value) as number[];
```
carysmills, line 58: This type was incorrect before

packages/polaris-viz/src/components/StackedAreaChart/utilities/yAxisMinMax.ts
```diff
@@ -21,5 +22,12 @@ export function yAxisMinMax(stackedValues: StackedSeries[]) {
     ),
   );
 
+  if (minY === Infinity || maxY === -Infinity) {
```
carysmills, line 25: @envex is this the handling we need here now that we don't default to Infinity internally?

susiekims (approved):
🎩 looks good to me! All the charts show the error state when one of the values is `Infinity` or `-Infinity`. Left some nit/questions that might be non-concerns.

PR #1743, opened 2024-10-23
Check chart bounds when mouse enters chart

envex:

We were only updating the dimensions on `resize`, so `dimensions.x` and `dimensions.y` were outdated when we were trying to calculate the tooltip position [here](https://github.com/Shopify/polaris-viz/blob/main/packages/polaris-viz/src/components/VerticalBarChart/Chart.tsx#L391-L392).

In this PR, we added a `onMouseEnter` and `onFocusIn`  handler to also update the dimensions when you interact with the chart.  

Fix https://github.com/Shopify/core-issues/issues/78973

Before:
https://github.com/user-attachments/assets/7f881726-54ef-45c1-aa6d-3a8634ced5d0

After:
https://github.com/user-attachments/assets/232f1c24-0355-4d1d-bfd0-8c30a6c47516

 
## Storybook link

Toggle the top/left margins and make sure the tooltips are in the correct position.
Test with scrolling as well. 

https://6062ad4a2d14cd0021539c1b-whnpeerjyh.chromatic.com/?path=/story/polaris-viz-charts-barchart-playground--external-tooltip-with-frame

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx
+++ b/packages/polaris-viz/src/components/BarChart/stories/playground/ExternalTooltip.stories.tsx
@@ -43,14 +43,22 @@ const Template: Story<BarChartProps> = (args: BarChartProps) => {
 
 const TemplateWithFrame: Story<BarChartProps> = (args: BarChartProps) => {
   const [ref, setRef] = useState<HTMLDivElement | null>(null);
+  const [height, setHeight] = useState(0);
+  const [marginLeft, setMarginLeft] = useState(0);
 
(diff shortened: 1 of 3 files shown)


carysmills (commented):
🎩 looks good to me

Can you just explain why we are making this change to ChartContainer and it now needs to care about recalculating the chart bounds on mouse enter, when it never did before? Just want to ensure this fixes the original problem. Should any of our previous code to handle portals be removed?

Does anything else need to happen to fix the area chart tooltip issue, or is this it?

carysmills (approved):

PR #1749, opened 2024-11-06
Change font size for touch devices

envex:

Resolves https://github.com/Shopify/core-issues/issues/80051

Change all the font-sizes to `14px` when on a mobile/touch device. This applies to legends, labels and values for all charts. 

**Desktop**

image

**Mobile/Touch**

image

## Storybook link

https://6062ad4a2d14cd0021539c1b-xwoqqwlwwf.chromatic.com/

### Before merging

### DIFF
--- a/packages/polaris-viz-core/src/constants.ts
+++ b/packages/polaris-viz-core/src/constants.ts
@@ -9,6 +9,8 @@ import {InternalChartType, ChartState, Hue} from './types';
 export const LINE_HEIGHT = 14;
 export const SMALL_CHART_HEIGHT = 125;
 export const FONT_SIZE = 11;
+export const TOUCH_FONT_SIZE = 14;
+
 export const FONT_FAMILY =
   'Inter, -apple-system, "system-ui", "San Francisco", "Segoe UI", Roboto, "Helvetica Neue", sans-serif';
 
--- a/packages/polaris-viz/src/components/Labels/utilities/truncateLabels.ts
+++ b/packages/polaris-viz/src/components/Labels/utilities/truncateLabels.ts
@@ -1,6 +1,6 @@
 import type {CharacterWidths} from '@shopify/polaris-viz-core';
-import {estimateStringWidth} from '@shopify/polaris-viz-core';
 
+import {estimateStringWidthWithOffset} from '../../../utilities';
 import type {PreparedLabels} from '../../../types';
 
 import {endLineTruncate} from './endLineTruncate';
@@ -23,7 +23,7 @@ export function truncateLabels({
 }: Props) {
   const truncatedLabels = [...labels];
 
-  labels.forEach((_, index) => {
+  labels.forEach(({fontSize}, index) => {
     truncatedLabels[index].truncatedWords = [];
     truncatedLabels[index].words = [];
 
@@ -35,7 +35,7 @@ export function truncateLabels({
       truncatedLabels[index].truncatedWords.push(truncatedLabels[index].text);
     } else {
       words.forEach((word) => {
-        const wordWidth = estimateStringWidth(word, characterWidths);
+        const wordWidth = estimateStringWidthWithOffset(word, fontSize);
 
         truncatedLabels[index].words.push({
           word,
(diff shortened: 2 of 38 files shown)


packages/polaris-viz/src/utilities/getFontSize.ts
```diff
@@ -0,0 +1,9 @@
+import {FONT_SIZE, MOBILE_FONT_SIZE} from '@shopify/polaris-viz-core';
+
+export function getFontSize() {
+  const isMobile = 'ontouchstart' in window || navigator.maxTouchPoints > 0;
```
carysmills, line 4: For some of the other mobile changes, we might watch to separate the mobile check into its own util/hook, separate from the font size but it doesn't need to be on this PR

packages/polaris-viz-core/src/constants.ts
```diff
@@ -9,6 +9,8 @@ import {InternalChartType, ChartState, Hue} from './types';
 export const LINE_HEIGHT = 14;
 export const SMALL_CHART_HEIGHT = 125;
 export const FONT_SIZE = 11;
+export const MOBILE_FONT_SIZE = 14;
```
carysmills, line 12: Should we call this touch device size, since it would also apply to tablets, with the way the check currently works?

carysmills (commented):
Generally looking good!

Should we be updating _all_ the font sizes as we currently are? My understanding was that we were just changing the axis labels. The legend text changes, for example, look a bit out of place to me, at least when compared to the hidden text.

Screenshot 2024-11-07 at 11 14 54 AM

If we do want the annotation sizes to change, I think there's something up with their truncation at this size
Screenshot 2024-11-07 at 11 20 45 AM

packages/polaris-viz/src/components/Labels/utilities/truncateLabels.ts
```diff
@@ -35,7 +35,7 @@ export function truncateLabels({
       truncatedLabels[index].truncatedWords.push(truncatedLabels[index].text);
     } else {
       words.forEach((word) => {
-        const wordWidth = estimateStringWidth(word, characterWidths);
+        const wordWidth = estimateStringWidthWithOffset(word, fontSize);
```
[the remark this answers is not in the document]
envex: We have the main character widths json file that has the widths based on `12px` font size.

We also have a [`character-width-offsets.json`](https://github.com/Shopify/polaris-viz/blob/815f565a816c724ef00b5db783af77b80c96c0be/packages/polaris-viz/src/data/character-width-offsets.json) file that contains a value for each font size from `0px` to `40px`.

So for example `14px` has an offset of `1.15` so `estimateStringWidthWithOffset` takes the size and multiplies by `1.15` to get the width of the string when the font-size is `14px`.

Does that rambling make a little sense?

envex:
@carysmills Removed the annotation font-size and fixed the `+ 3 more` font size on legends.

packages/polaris-viz/src/components/Labels/utilities/truncateLabels.ts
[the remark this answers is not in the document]
susiekims: Wow TIL I've seen that util around but never knew what it meant by `WithOffset` until now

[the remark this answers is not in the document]
carysmills: I think so! Just wanted to ensure we are taking into account the character widths still, which it sounds like we are!

carysmills (commented):
This seems like an unexpected change, as far as I can tell. What do you think? On this story ?path=/docs/polaris-viz-charts-linechart--series-colors-up-to-sixteen

| Before    | After |
| -------- | ------- |
| Screenshot 2024-11-08 at 9 11 55 AM  | Screenshot 2024-11-08 at 9 12 14 AM   |

That was something I noticed on desktop. Looking on touch devices now

carysmills (commented):
Touch looked good! Just the issue noted above

envex:
@carysmills I'll dig in, but it's likely the new sizes are being too aggressive on the "Drop labels that touch" logic.

I vaguely remember fixing an issue here as part of this PR as well, so maybe we just need to be less aggressive .

packages/polaris-viz/src/hooks/useLinearLabelsAndDimensions.ts
```diff
@@ -55,9 +50,8 @@ export function useLinearLabelsAndDimensions({
     return clamp({
       amount: longestLabelWidth,
       min: 0,
-      max: MAX_LINEAR_LABEL_WIDTH,
```
susiekims, line 58: I think the visual change @carysmills noticed was because we removed this line

envex: Yup! Going to revert that change.

PR #1750, opened 2024-11-06
Set LineChart Tooltip at static Y when on touche device

envex:

When on a touch device we want the `LineChart` tooltip to always display above the chart instead of under the users finger.

### Before merging


packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -243,9 +251,11 @@ export function Chart({
         max: data[longestSeriesIndex].data.length - 1,
       });
 
+      const {x, y} = getXYFromEventType(event);
```
carysmills, line 254: At some point, not necessarily now, I think it would be nice to consolidate some of the tooltip positioning logic. I'm mostly thinking of what's in this function, what's in `getAlteredPosition` and what's in the actual tooltip component and whether there's a way to simplify it all a bit

packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -226,10 +227,17 @@ export function Chart({
     index,
     eventType,
   }: TooltipPositionParams): TooltipPosition {
+    event?.preventDefault();
```
carysmills, line 230: We want to do this on every device type?

carysmills (commented):
This generally feels really good to me 🎉 I would like to try the scrolling part in the context of the admin, but I think the tooltip changes work really well. We'd probably want the exact same for area chart and I am not super sure what the behaviour should be for bar charts, probably about the same?

PR #1756, opened 2024-11-18
Envex/mobile tooltip

envex:
### Before merging


carysmills:
Replaced myself as a reviewer with @philschoefer 👋 
let me know if I am needed!

philschoefer:
Thanks @envex! Some 🎩 ing feedback on things I noticed:

1. I'm noticing some `xAxis` and `yAxis` strings that I don't see on the regular polaris viz storybook: image
2. It seems like the positioning of the active data point is just off to the side of my finger which feels a bit weird. https://github.com/user-attachments/assets/22fb9384-413c-469f-828c-9e4e5d39f0c7
3. The positioning of the tooltip seems a bit confusing. From the issue I understand they want it to be more like home where the tooltip is fixed, but it's fixed in the center above the active data point which seems more intuitive then the top left positioning we have now IMO. Outside of where it should be, I think it would make sense if we do it the same way just for consistency sake: https://github.com/user-attachments/assets/23a36c6f-1118-4193-82bd-f65693aff797

FWIW: I played around with line charts on Apple Health a bit and they do something similar.

https://github.com/user-attachments/assets/e96e2e51-482a-4e00-b84b-e85214b8b57f

envex:
@philschoefer 

Which story are you seeing the wrong axis labels for point 1?

2 and 3 have been addressed.

susiekims (commented):
Looks like the logic we added [here](https://github.com/Shopify/polaris-viz/pull/1743) is behaving strangely with the long press 🤔 The container bounds are not being updated, we may need to add a `onTouchStart` handler as well as the `onMouseEnter`

PR #1757, opened 2024-11-19
Rename dimensions to containerBounds

envex:

> [!NOTE]  
> I'd like to merge this before https://github.com/Shopify/polaris-viz/pull/1755 & https://github.com/Shopify/polaris-viz/pull/1756

We're using `dimensions` and `bounds` all over the repo, but along the way sometimes a `bounds` prop will only contain dimensions and vice versa.

This PR renames the `dimension` prop to `containerBounds` so we can easily distinguish between what's a container value and whats a chart value.

image

In this example, the red bounds are considered `containerBounds`. This gives us the `height` and `width` as well as the `x/y` position of the container on the page.

We also have a concept of `chartBounds`, which is the `height/width/x/y` of the red outline. This is the bounds of the chart itself relative to the container.

There should be no visual changes.

## Storybook link

https://6062ad4a2d14cd0021539c1b-kjvrjvketz.chromatic.com/

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/LegendContainer/components/HiddenLegendTooltip.tsx
+++ b/packages/polaris-viz/src/components/LegendContainer/components/HiddenLegendTooltip.tsx
@@ -13,7 +13,7 @@ import {
   useChartContext,
   useTheme,
 } from '@shopify/polaris-viz-core';
-import type {BoundingRect, LabelFormatter} from '@shopify/polaris-viz-core';
+import type {Dimensions, LabelFormatter} from '@shopify/polaris-viz-core';
 
 import {getFontSize} from '../../../utilities/getFontSize';
 import type {LegendData} from '../../../types';
@@ -29,12 +29,12 @@ import style from './HiddenLegendTooltip.scss';
 interface Props {
   activeIndex: number;
   colorVisionType: string;
+  containerDimensions: Dimensions;
   data: LegendData[];
   label: ReactNode;
   setActivatorWidth: (width: number) => void;
   theme?: string;
   lastVisibleIndex?: number;
-  dimensions?: BoundingRect;
   seriesNameFormatter?: LabelFormatter;
 }
 
@@ -44,12 +44,12 @@ export const LEGEND_TOOLIP_Z_INDEX = 520;
 export function HiddenLegendTooltip({
   activeIndex,
   colorVisionType,
+  containerDimensions,
   data,
   theme,
   label,
   lastVisibleIndex = 0,
   setActivatorWidth,
-  dimensions,
   seriesNameFormatter,
 }: Props) {
   const selectedTheme = useTheme();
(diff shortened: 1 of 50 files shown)


packages/polaris-viz/src/components/LegendContainer/components/HiddenLegendTooltip.tsx
```diff
@@ -59,7 +57,7 @@ export function HiddenLegendTooltip({
   const container = useRootContainer(tooltipId);
   const tooltipRef = useRef<HTMLDivElement>(null);
   const activatorRef = useRef<HTMLButtonElement>(null);
-  useColorVisionEvents({enabled: true, root: LEGEND_TOOLTIP_ID, dimensions});
+  useColorVisionEvents({enabled: true, root: LEGEND_TOOLTIP_ID});
```
carysmills, line 60: Why are we no longer passing dimensions here?

packages/polaris-viz/src/components/LineChart/Chart.tsx
```diff
@@ -126,14 +125,19 @@ export function Chart({
...
+
   const {legend, setLegendDimensions, height, width} = useLegend({
     data: [
       {
         shape: 'Line',
         series: data,
       },
     ],
-    dimensions,
+    dimensions: containerDimensions,
```
carysmills, line 140: This line used to be dimensions and so did line 261, so I am a bit confused why they are replaced with different variables, not the same one. Is this intentional? Were we using the wrong variable before?

carysmills (commented):
Are the height and width of `containerBounds` and `chartBounds` sometimes different? If they are the same, I am wondering if we can just have height and width associated with one of them, to avoid confusion.

Also just going back to the original problem:
> We're using dimensions and bounds all over the repo, but along the way sometimes a bounds prop will only contain dimensions and vice versa.

Would it possibly make sense to just ensure that one prop always has all the dimensions we are looking for, rather than having two props that are similar?

packages/polaris-viz/src/components/LineChart/Chart.tsx
envex: This is intentional. This was using the original `dimensions` prop that was actually bounds.

Now `containerDimensions` only passed `height` and `width`.

envex:
> Are the height and width of containerBounds and chartBounds sometimes different? If they are the same, I am wondering if we can just have height and width associated with one of them, to avoid confusion.

The `containerBounds` and `chartBounds` will almost always be different.

`containerBounds`: `x/y` are relative to the browser. `height/width` are the entire size of the container.
`chartBounds`: `x/y` are relative to the container and take into account the theme padding. `height/width` are the drawable area of the chart, so mainly `container width - yAxis labels`.

packages/polaris-viz/src/components/ChartContainer/components/ChartDimensions/ChartDimensions.tsx
susiekims, line 1: Would it make sense to rename is to `ContainerBounds` to follow the new naming pattern?

envex:
@susiekims I renamed the props and `ChartDimenions` component based on your feedback.

envex:
I hate myself for doing this, but I wonder if we even need `containerDimensions` to be it's own thing.

Most places we should be able to just pass down `containerBounds` even though we only need the `height/width`.

Thoughts?

susiekims:
> I hate myself for doing this, but I wonder if we even need `containerDimensions` to be it's own thing.
> 
> Most places we should be able to just pass down `containerBounds` even though we only need the `height/width`.
> 
> Thoughts?


@envex I was thinking about that actually! We don't really need `chartDimensions` _and_ `chartBounds`, or `containerDimensions` _and_ `containerBounds`, since `bounds` has `dimensions`. Maybe we could even store it chart context so we don't have to prop drill it down? 

That being said, I'm happy to let this PR go in first and come back to it since it's still a step in a right direction and will make that refactor way easier.

PR #1760, opened 2024-11-19
Debarrel packages/polaris-viz

envex:
### Before merging


carysmills (commented):
Not urgent but when you're able to, can you add some info to the description about why we are making this change, any files in particular to focus on, or any commands we should try running to make sure things are working okay

PR #1761, opened 2024-11-19
fix, Tooltip should be render inside `polaris_viz_tooltip_root`

Passanelli:

We need to make use of the `polaris_viz_tooltip_root` element through `createPortal` in order to make tooltips always visible

## Storybook link

storybook [URL](https://6062ad4a2d14cd0021539c1b-ugedechvhc.chromatic.com/?path=/story/polaris-viz-charts-grid--default)

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/Grid/Grid.tsx
+++ b/packages/polaris-viz/src/components/Grid/Grid.tsx
@@ -1,15 +1,9 @@
-import React, {
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import type {LabelFormatter} from '@shopify/polaris-viz-core';
 import {
   DEFAULT_CHART_PROPS,
   useChartPositions,
+  useUniqueId,
 } from '@shopify/polaris-viz-core';
 import {scaleLinear} from 'd3-scale';
 
@@ -26,7 +20,6 @@ import {
   TOOLTIP_WIDTH,
   TOOLTIP_HEIGHT,
   TOOLTIP_HORIZONTAL_OFFSET,
-  TOOLTIP_VERTICAL_OFFSET,
   Y_AXIS_LABEL_WIDTH,
   X_AXIS_HEIGHT,
   DEFAULT_GROUP_COLOR,
@@ -61,15 +54,6 @@ export function Grid(props: GridProps) {
   const [isSmallContainer, setIsSmallContainer] = useState(false);
   const [groupSelected, setGroupSelected] = useState<CellGroup | null>(null);
 
-  const tooltipRef = useRef<HTMLDivElement>(null);
-  const [tooltipDimensions, setTooltipDimensions] = useState<{
-    width: number;
-    height: number;
-  }>({
-    width: TOOLTIP_WIDTH,
-    height: TOOLTIP_HEIGHT,
-  });
-
   const {
     cellGroups = [],
     xAxisOptions = {},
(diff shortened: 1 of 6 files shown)


packages/polaris-viz/src/components/Grid/components/Tooltip.tsx
```diff
@@ -1,54 +1,68 @@
+import {forwardRef} from 'react';
+import {createPortal} from 'react-dom';
+
 import type {CellGroup} from '../types';
+import {useRootContainer} from '../../../hooks/useRootContainer';
 
 import styles from './Tooltip.scss';
 
+const TOOLTIP_ID = 'polaris_viz_tooltip_root';
```
envex, line 9: Can we extract this out into a shared constant that `packages/polaris-viz/src/components/TooltipWrapper/TooltipWrapper.tsx` can also use?

packages/polaris-viz/src/constants.ts
```diff
@@ -62,3 +62,4 @@ export const WARN_FOR_DEVELOPMENT = IS_DEVELOPMENT && !IS_TEST;
 export const HOVER_TARGET_ZONE = 48;
 export const CROSSHAIR_ID = 'Crosshair';
 export const DEFAULT_ANIMATION_DELAY = 100;
+export const TOOLTIP_ID = 'polaris_viz_tooltip_root';
```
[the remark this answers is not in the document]
envex: Oh, I'm dumb. Sorry :|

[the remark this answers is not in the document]
Passanelli: haha not problem at all!

packages/polaris-viz/src/components/Grid/components/Tooltip.tsx
```diff
@@ -8,47 +13,55 @@ interface TooltipProps {
...
-            >
-              <path
-                strokeLinecap="round"
-                strokeLinejoin="round"
-                d="M13.5 4.5 21 12m0 0-7.5 7.5M21 12H3"
-              />
-            </svg>
-            <p className={styles.GroupGoal}>{group.goal}</p>
+    return createPortal(
+      !group ? null : (
```
envex, line 21: Are we able to bail out early and not even call `createPortal`?

Sorry, it's been a while since I implemented it so this question is more being curious.

Passanelli: oh maybe we can i will double check for sure and update it if works!

packages/polaris-viz/src/components/Grid/Grid.tsx
```diff
@@ -89,6 +105,13 @@ export function Grid(props: GridProps) {
     return new Set([group.id, ...(group.connectedGroups ?? [])]);
   };
 
+  useLayoutEffect(() => {
```
mollerjorge, line 108: I think we can get rid of this and use the `TOOLTIP_MIN_HEIGHT` constants after we rename `TOOLTIP_VERTICAL_OFFSET` into `TOOLTIP_MIN_HEIGHT`

packages/polaris-viz/src/components/Grid/Grid.tsx
```diff
@@ -105,21 +128,17 @@ export function Grid(props: GridProps) {
...
-        y = rect.bottom - containerRect.top + TOOLTIP_HORIZONTAL_OFFSET;
+      } else if (bottomSpace >= tooltipDimensions.height) {
+        x = rect.left;
+        y = rect.bottom + TOOLTIP_HORIZONTAL_OFFSET;
         placement = 'bottom';
       } else {
-        x = rect.left - containerRect.left;
-        y = rect.top - containerRect.top - TOOLTIP_VERTICAL_OFFSET;
+        x = rect.left;
+        y = rect.top - TOOLTIP_VERTICAL_OFFSET;
```
mollerjorge, line 141: maybe renaming this to `TOOLTIP_MIN_HEIGHT` gives more meaning

packages/polaris-viz/src/components/Grid/Grid.tsx
```diff
@@ -323,13 +342,12 @@ export function Grid(props: GridProps) {
...
-      {tooltipInfo && (
-        <Tooltip
-          x={tooltipInfo.x}
-          y={tooltipInfo.y}
-          group={groupSelected || hoveredGroup}
-        />
-      )}
+      <Tooltip
+        ref={tooltipRef}
+        x={tooltipInfo?.x ?? 0}
```
mollerjorge, line 347: I think we can remove the ?? and bring back the `tooltipInfo &&` I think we only removed it when trying to calculate the height

PR #1767, opened 2024-12-02
Do not show ellipsis if only one character is truncated

carysmills:

Addresses feedback about not using an ellipsis to truncate if we are only removing one character

For the example pictured below, `truncateSingleLine` was what needed to be updated. I figured it made sense to update the other instances where we add an ellipsis, but I am a bit unclear on what circumstances these are actually used under. @envex any insight here?

I don't think an issue was opened for this

| Before  | After  |
|---|---|
|   |   |

## Storybook link
https://6062ad4a2d14cd0021539c1b-gbmlhdxtss.chromatic.com/?path=/story/polaris-viz-charts-barchart-playground--vertical-truncate

### Before merging

### DIFF
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Do not show ellipsis if only one character is truncated 
 - Fixed an issue with the `<StackedAreaChart />` that would cause the chart to break when series length increases
 
 ### Changed


envex (approved):
Looks good to me and all makes sense.

You're right, it makes sense to do it wherever we truncate. I think the multiline truncation is pretty rare anyways, so we're good.

packages/polaris-viz/src/components/Labels/utilities/endWordTruncate.ts
```diff
@@ -30,5 +30,9 @@ export function endWordTruncate({word, targetWidth, characterWidths}: Props) {
     newWord = `${newWord}${ELLIPSIS}`;
   }
 
+  if (newWord.length === word.length - 1) {
```
nickpresta, line 33: I don't know if it's worth introducing this complexity (yet, or ever), but how do we want this to handle unicode characters where a single visible character may not correspond to a length of `1`:

```javascript
function getGraphemeCount(str) {
  const segmenter = new Intl.Segmenter("en-US", { granularity: "grapheme" });
  return [...segmenter.segment(str)].length;
}

function getCodePointsCount(str) {
  return [...str].length;
}

console.log(getGraphemeCount("안녕하세요")); // 5
console.log(getCodePointsCount("안녕하세요")); // 5
console.log("안녕하세요".length); // 5

console.log(getGraphemeCount("👨‍👩‍👧‍👧")); // 1
console.log(getCodePointsCount("👨‍👩‍👧‍👧")); // 7
console.log("👨‍👩‍👧‍👧".length); // 11
```

carysmills: Interesting! I have seen emojis in product titles so this is probably actually a problem. I will open an issue

PR #1771, opened 2024-12-02
Fix issue where legends would get formatted twice

envex:

We were running `seriesNameFormatter` multiple times on a `<Legend />` in the `<DonutChart />` component.

This is causing issues in web because the `i18n.translate` function is running the second time on a formatted value, which throws a bugsnag error.

| Before  | After  |
|---|---|
|image|image|

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/SparkLineChart/tests/Chart.test.tsx
+++ b/packages/polaris-viz/src/components/SparkLineChart/tests/Chart.test.tsx
@@ -26,7 +26,7 @@ describe('<Chart />', () => {
     const offsetLeft = 100;
     const offsetRight = 50;
     const margin = 2;
-    const mockWidth = 0;
+    const mockWidth = 600;
 
     mount(
       <Chart
--- a/packages/polaris-viz/CHANGELOG.md
+++ b/packages/polaris-viz/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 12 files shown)


packages/polaris-viz/src/components/SparkLineChart/tests/Chart.test.tsx
```diff
@@ -26,7 +26,7 @@ describe('<Chart />', () => {
     const offsetLeft = 100;
     const offsetRight = 50;
     const margin = 2;
-    const mockWidth = 0;
+    const mockWidth = 600;
```
carysmills, line 29: Confused why we have Sparkline test changes on this PR. Would be good to add a comment here to explain or in the PR description in the future.

carysmills (commented):
It's a bit hard to understand the changes on this PR because there seem to be some other testing changes mixed in. In the future, let's try to keep the changes in scope to the issue you are fixing, or explain why the changes were needed in other components/tests. This will help reviewers.

Could you please explain when this bug occurred and how I can recreate it and see that it is fixed now? Looking here https://polaris-viz.shopify.dev/?path=/story/polaris-viz-charts-donutchart--formatted-values, a `seriesNameFormatter` was being passed in but I do not see it being applied twice.

envex:
@carysmills When I was trying to add a test to the `DonutChat.test.tsx` I noticed that the tests were actually all broken, but still passing correctly - that's the reason there's other test changes. Fixing the broken `DonutChart` test revealed a few other places that tests weren't working correctly.

packages/polaris-viz/src/components/SparkLineChart/tests/Chart.test.tsx
envex: Adding a little explanation here: https://github.com/Shopify/polaris-viz/pull/1771#issuecomment-2514953274

PR #1777, opened 2024-12-09
Bar chart labels to partially display like Linear charts

markmssd:

With Linear charts, the X-Axis displays labels nicely even if the width does not fit them all. It smartly displays less and less labels as it can, until it can no more show any of them.

However for Bar charts, it behaves differently. If all the labels can't fit within the width, it simply displays none of them. With this PR, it changes this behaviour to be similar to the Linear charts.

I have created a new story for Bar Charts, `Monthly Reports`, to show an example with a typical use case, showing a 30-days monthly report.

No

| Before  | After  |
|---|---|
|   before | after  |

 
### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/BarChart/stories/MonthlyReport.stories.tsx
+++ b/packages/polaris-viz/src/components/BarChart/stories/MonthlyReport.stories.tsx
@@ -1,0 +1,18 @@
+import type {Story} from '@storybook/react';
+
+export {META as default} from './meta';
+
+import type {BarChartProps} from '../../../components';
+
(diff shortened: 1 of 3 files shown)


markmssd:
Closing as it's not the right approach, will open an issue instead.

envex:
We likely won't merge this because we don't want to remove labels in `BarChart`. Unlike `LineChart`, where the assumption of linearity allows us to drop labels without a 1:1 relationship to data points,`BarChart` require each bar to have its own label. If you are working with linear data, consider using a `LineChart` instead.

envex:
@carysmills Any thoughts on making this a prop?

Issue #1787, opened 2025-01-03
Part of the tooltip title can be hidden or overflow.


AnhVu23:
## Bug summary

In the bar chart, the tooltip title, which is long and does not have any spaces, can be hidden or overflow.

## Expected behavior

The tooltip title should either display all the content, or break into new line.

## Actual behavior

The tooltip can either be overflow or hidden if the string is too long and does not have any spaces.

## Steps to reproduce the problem

Check this codesandbox: https://codesandbox.io/p/sandbox/shopify-polaris-viz-react-typescript-playground-forked-3xsxy6?file=%2Fsrc%2Findex.tsx%3A12%2C13

## Reduced test case

https://codesandbox.io/p/sandbox/shopify-polaris-viz-react-typescript-playground-forked-3xsxy6?file=%2Fsrc%2Findex.tsx%3A12%2C13

## Specifications

- Polaris-Viz version number: 15.3.2
- Browser: Chrome
- Device: MacBook Pro
- Operating System: Sequoia 15.1.1

PR #1789, opened 2025-01-08
Fix tooltip overflow when displaying long title without space

AnhVu23:

When displaying long tooltip title without space, the text is overflow. More details and the bug reproduction are mentioned here: https://github.com/Shopify/polaris-viz/issues/1787.

PR aims to fix the issue by adding **overflow-wrap: break-word** to the `<TooltipTitle />` component.

Issue link: https://github.com/Shopify/polaris-viz/issues/1787.

Before 

Screenshot 2025-01-08 at 13 04 00

After

Screenshot 2025-01-08 at 13 03 33

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/TooltipContent/stories/LongTitleWithoutSpace.stories.tsx
+++ b/packages/polaris-viz/src/components/TooltipContent/stories/LongTitleWithoutSpace.stories.tsx
@@ -1,0 +1,33 @@
+import {DARK_THEME} from '@shopify/polaris-viz-core';
+import type {Story} from '@storybook/react';
+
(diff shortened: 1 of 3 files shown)


packages/polaris-viz/src/components/TooltipContent/stories/LongTitleWithoutSpace.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import {DARK_THEME} from '@shopify/polaris-viz-core';
```
envex, line 1: We probably don't need this story anymore, would you mind removing it?

PR #1791, opened 2025-01-14
Add tooltips to Funnel Chart Next

michaelnesen:

This PR adds back the funnel chart tooltips removed in https://github.com/Shopify/polaris-viz/pull/1785, and makes some changes to when we show the `dropped off` labels. The first segment will now only show a `Reached this step` label, with subsequent tooltips containing the `dropped off` data related to the previous segment. 

Screenshot 2025-01-14 at 5 20 56 PM

 
## Storybook link

https://6062ad4a2d14cd0021539c1b-gqxlmulfcd.chromatic.com/?path=/story/polaris-viz-charts-funnelchartnext--default

- Test out `showTooltip` prop and ensure it works
- Ensure tabbing through chart works well

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
+++ b/packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
@@ -1,7 +1,11 @@
 import type {ReactNode} from 'react';
 import {Fragment, useMemo, useState} from 'react';
 import {scaleBand, scaleLinear} from 'd3-scale';
-import type {DataSeries, LabelFormatter} from '@shopify/polaris-viz-core';
+import type {
+  DataPoint,
+  DataSeries,
+  LabelFormatter,
+} from '@shopify/polaris-viz-core';
 import {
   uniqueId,
   LinearGradientWithStops,
@@ -33,11 +37,14 @@ import {
   TOOLTIP_HEIGHT,
   TOOLTIP_WIDTH,
 } from './constants';
-import type {FunnelChartNextProps} from './FunnelChartNext';
 
 export interface ChartProps {
   data: DataSeries[];
-  tooltipLabels: FunnelChartNextProps['tooltipLabels'];
+  tooltipLabels: {
+    reached: string;
+    dropped: string;
+  };
+  showTooltip?: boolean;
   seriesNameFormatter: LabelFormatter;
   labelFormatter: LabelFormatter;
   percentageFormatter?: (value: number) => string;
@@ -47,6 +54,7 @@ export interface ChartProps {
 export function Chart({
   data,
   tooltipLabels,
+  showTooltip = true,
   seriesNameFormatter,
   labelFormatter,
   percentageFormatter = (value: number) => {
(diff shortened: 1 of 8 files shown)


packages/polaris-viz/src/components/FunnelChartNext/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  return (
+    <TooltipContentContainer
+      maxWidth={TOOLTIP_WIDTH}
+      theme={DEFAULT_THEME_NAME}
+    >
+      {() => (
+        <Fragment>
+          <TooltipTitle theme={DEFAULT_THEME_NAME}>{point.key}</TooltipTitle>
```
mollerjorge, line 70: Can we add some `aria-labels` in the title and the data row to make it accessible through screen readers? 

e.g. something like this for the data row: ```aria-label={`${key}: ${value}, ${percentageFormatter(percent)}`}```

packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
```diff
@@ -182,7 +210,55 @@ export function Chart({
             </Fragment>
           );
         })}
+        <TooltipWithPortal>{getTooltipMarkup()}</TooltipWithPortal>
       </g>
     </ChartElements.Svg>
   );
+
+  function getTooltipMarkup() {
+    if (tooltipIndex == null) {
```
mollerjorge, line 219: maybe we can use an early return here to make things a bit more clear like:

```
  if (tooltipIndex == null || !dataSeries[tooltipIndex]) {
    return null;
  }

  const activeDataSeries = dataSeries[tooltipIndex];
  const xPosition = calculateXPosition(activeDataSeries);
  const yPosition = calculateYPosition(activeDataSeries);

  return (
    <FunnelTooltip x={xPosition} y={yPosition}>
      <Tooltip
        activeIndex={tooltipIndex}
        dataSeries={dataSeries}
        tooltipLabels={tooltipLabels}
        labelFormatter={labelFormatter}
        percentageFormatter={percentageFormatter}
      />
    </FunnelTooltip>
  );

```

packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
```diff
@@ -16,18 +16,28 @@ import {
...
   SEGMENT_WIDTH_RATIO,
+  TOOLTIP_HORIZONTAL_OFFSET,
+  TOOLTIP_HEIGHT,
+  TOOLTIP_WIDTH,
 } from './constants';
+import type {FunnelChartNextProps} from './FunnelChartNext';
 
 export interface ChartProps {
   data: DataSeries[];
+  tooltipLabels: FunnelChartNextProps['tooltipLabels'];
```
mollerjorge, line 40: I wonder if we should make this `tooltipLabels` optional as we are not going to use them for `open` funnels?

mollerjorge (approved):

PR #1792, opened 2025-01-16
Fix percentile index for line chart relational

mollerjorge:

We have two different logics for calculating the `percentileIndex` for `LineChartRelational`. This was causing a mismatch on indexes when trying to highlight a path in the chart from the legend item.

The solution is to unify this with the same logic so that there is always a match.

https://github.com/user-attachments/assets/0a2dbf5e-74c6-4f3a-ab4d-0003b2ec2ae5


michaelnesen (approved):
Great find! Thanks for fixing 🙌

PR #1799, opened 2025-02-04
WIP: Simplify Tooltips

envex:
### Before merging


packages/polaris-viz/src/components/TooltipWrapper/hooks/useRepositionTooltip.ts
```diff
@@ -0,0 +1,337 @@
...
+
+  const containerMax =
+    containerWidth - tooltipDimensions.width - SCROLLBAR_SIZE;
+  let newX = x;
+
+  // If the tooltip has been repositioned where it would render
+  // outside the browser window, or cover the data series,
+  // we want to reposition it again around the series bounds.
+  if (isOutsideWindow && seriesBounds != null) {
+    const positions: Position[] = [
```
envex, line 142: For charts that don't update their position, we likely pass a `seriesBounds` prop.

This prop is a set of `x, y, height and width` values that we use to position the tooltip around.

For example, in a BarChart, if the tooltip has been repositioned into the `seriesBounds` we will try and reposition the tooltip to the left of the `seriesBounds` so that the tooltip does not cover the bars.

PR #1805, opened 2025-02-12
Add -50 font-weights

envex:

`TrendIndicators` that rendered large values were getting cut off because we default to a font-weight of `650` but we didn't support that weight in the `character-width-offsets`.

image

| Before  | After  |
|---|---|
|image|image|
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-xmtyzhmtdp.chromatic.com/?path=/story/chromatic-components--trendindicator

### Before merging


michaelnesen (approved):
🫡

PR #1806, opened 2025-02-12
Remove inactive tooltip rows

envex:

To clean up the visual noise of the tooltips, we're removing inactive items from rendering in the tooltip when other series have focus.

| Before  | After  |
|---|---|
|image|image|

 
### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/LineChartPredictive/utilities/renderLinearPredictiveTooltipContent.tsx
+++ b/packages/polaris-viz/src/components/LineChartPredictive/utilities/renderLinearPredictiveTooltipContent.tsx
@@ -57,11 +57,17 @@ export function renderLinearPredictiveTooltipContent(
       const isPredictiveStartKey = metadata?.startKey === activeKey;
       const isHidden = isNull || isPredictiveStartKey;
 
+      if (
+        isComparison !== true &&
+        activeColorVisionIndex !== -1 &&
+        index !== activeColorVisionIndex
+      ) {
+        return null;
+      }
+
       return (
         <TooltipRow
-          activeIndex={activeColorVisionIndex}
           color={color}
-          index={index}
           isHidden={isHidden}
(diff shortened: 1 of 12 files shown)


packages/polaris-viz/src/components/LineChartPredictive/utilities/renderLinearPredictiveTooltipContent.tsx
```diff
@@ -57,11 +57,17 @@ export function renderLinearPredictiveTooltipContent(
       const isPredictiveStartKey = metadata?.startKey === activeKey;
       const isHidden = isNull || isPredictiveStartKey;
 
+      if (
+        isComparison !== true &&
+        activeColorVisionIndex !== -1 &&
+        index !== activeColorVisionIndex
+      ) {
+        return null;
+      }
```
michaelnesen, line 66: More of a UX question, but I noticed that displaying all the series in the tooltip when there's no active index—and then switching to just one when you hover over a series can create some jarring behaviour, especially with higher series lengths. It keeps flipping between the two states and looks a bit off. Do you think there's something we could add to help minimize that effect?


https://github.com/user-attachments/assets/d79ba06d-30bc-4202-ac3e-c504beaa5c9d

envex: @adamperron Thoughts?

adamperron: I need some time to think about this one

envex:
@susiekims I've fixed your issue and @michaelnesen we've added a slight delay when resetting the tooltip content so it shouldn't jump around as much in that case.

michaelnesen (approved):
Nice! Changes look good to me 👍

PR #1808, opened 2025-02-13
Add support for multiple comparison series to LineChart

envex:

As part of the [ongoing tooltip work](https://github.com/Shopify/core-issues/issues/77473) we want to allow charts to display multiple comparison lines.

These lines should be hidden until the matching series line is focused. If a chart only has a single comparison series, it show both the series and comparison series.

Resolves https://github.com/Shopify/core-issues/issues/84724

https://github.com/user-attachments/assets/8e2f89eb-1b30-4ce1-9cd6-64ba606b0e5f
 
### Before merging


packages/polaris-viz/src/components/TooltipContent/TooltipContent.tsx
```diff
@@ -61,13 +63,25 @@ export function TooltipContent({
                     seriesIndex,
                   ) => {
                     if (
+                      // TODO: Check this
```
susiekims, line 66: Seems right? At least from tophatting 🤔

packages/polaris-viz/src/components/TooltipContent/TooltipContent.tsx
```diff
@@ -61,13 +63,25 @@ export function TooltipContent({
...
 
+                    if (
+                      isComparisonSeriesHidden({
+                        index: seriesIndex,
+                        activeColorVisionIndex,
+                        comparisonSeriesIndexes,
+                      })
+                    ) {
+                      return null;
+                    }
```
susiekims, line 83: - If there's more than 2 data points in the series AND it's not a comparison AND is not active, return null
- If it has a corresponding comparison series that's hidden, return null
- Else, return the tooltip row

Is that right?

packages/polaris-viz/src/utilities/isComparisonSeriesHidden.ts
```diff
@@ -0,0 +1,24 @@
+import type {ComparisonSeriesIndex} from '@shopify/polaris-viz-core/src/types';
+
+export function isComparisonSeriesHidden({
```
susiekims, line 3: nit: maybe `hasHiddenComparisonSeries` would be clearer, to describe the `if (comparisonSeries == null) return false` case

susiekims (commented):
🎩 looks good, just a few nits/questions

michaelnesen (approved):
Going to approve as I am out next week and don't want to block 

One weird interaction but overall code and 🎩 are looking great 👍

packages/polaris-viz/src/components/TooltipContent/TooltipContent.tsx
envex: Ok, writing this out cause it's also kinda confusing for myself, but I can't think of a better solution that works.

For a chart with a single line and single comparison, we want to show all the tooltip rows when hovering either line.

**When not hovering a line**

- Show all lines

**When hovering a line**

- Show all lines

https://github.com/user-attachments/assets/2adf0e3b-97a0-4448-93b0-6d6a5601074b

but, if we're rendering multiple comparisons, when hovering a line we only want to show the current line and it's comparison line.......but if we're not hovering any lines, we only want to show non-comparison lines

**When not hovering a line**

- Show all non-comparison lines

**When hovering a line**

- Show current hovered line
- Show it's matching comparison line

https://github.com/user-attachments/assets/840707c0-3492-4dd7-b9eb-d697ce0d5d34

susiekims (approved):

PR #1809, opened 2025-02-20
Render TrendIndicator in tooltips

envex:

Adds the ability to render a `TrendIndicator` along side a data point in the tooltips. This should allow us to remove the `AnalyticsTooltip` need in `analytics-ui-components` and we can rely on using only the PV tooltip rendering.

image
 
## Storybook link

https://6062ad4a2d14cd0021539c1b-iwwvtwfwgf.chromatic.com/?path=/story/polaris-viz-charts-linechart--default

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/TooltipContent/components/TooltipRow/TooltipRow.tsx
+++ b/packages/polaris-viz/src/components/TooltipContent/components/TooltipRow/TooltipRow.tsx
@@ -1,36 +1,37 @@
 import {useTheme} from '@shopify/polaris-viz-core';
-import type {Shape, Color} from '@shopify/polaris-viz-core';
+import type {
+  Shape,
+  Color,
+  TrendIndicator as TrendIndicatorType,
+} from '@shopify/polaris-viz-core';
 
+import {TrendIndicator} from '../../../TrendIndicator';
 import {PREVIEW_ICON_SIZE} from '../../../../constants';
 import {SeriesIcon} from '../../../shared/SeriesIcon';
 import {TITLE_MARGIN} from '../../constants';
 
 import styles from './TooltipRow.scss';
 
-interface Props {
+export interface Props {
   label: string;
   shape: Shape;
   value: string;
   color?: Color;
   isComparison?: boolean;
-  isHidden?: boolean;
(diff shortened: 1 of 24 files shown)


packages/polaris-viz/src/components/TooltipContent/components/TooltipRow/TooltipRow.tsx
```diff
@@ -64,6 +66,20 @@ export function TooltipRow({
...
 }
+
+const DEFAULT_COLUMN_COUNT = 2;
+
+export function getColumnCount(data: TooltipData) {
+  const hasTrend = data.data.some(({trend}) => trend != null);
+  const hasColor = data.data.some(({color}) => color != null);
+
+  return DEFAULT_COLUMN_COUNT + (hasColor ? 1 : 0) + (hasTrend ? 1 : 0);
+}
```
susiekims, line 85: Would it make sense to define this in `TooltipSeries`? Or maybe a separate util file 🤔

packages/polaris-viz/src/components/TooltipContent/components/TooltipSeries/TooltipSeries.tsx
```diff
@@ -1,16 +1,31 @@
...
+
+import {TooltipSeriesName} from '../TooltipSeriesName';
 
 import styles from './TooltipSeries.scss';
 
-interface Props {
+export interface Props {
   children: ReactNode;
-  isEmpty?: boolean;
+  columnCount?: number;
```
susiekims, line 11: A `columnCount` prop feels a little too imperative, have you thought about making the props more declarative i.e. `hasPreview` and `hasTrend`? Then `TooltipSeries` can derive the column count based on those booleans

susiekims (commented):
Very handy! Could you add a trend indicator to one of the tooltip stories for easier tophatting

packages/polaris-viz/src/components/TooltipContent/components/TooltipRow/TooltipRow.tsx
envex: Moved this to it's own file!

PR #1811, opened 2025-02-25
Add provenance statement for polaris viz

carysmills:
Adds a mandatory [provenance statement](https://docs.npmjs.com/generating-provenance-statements) when the package is published. You can read more about this requirement in #proj-npm-security-update. The deadline is March 20 but trying to get this done ASAP in case it requires a larger update or something unexpected.

The straightforward way to do this would be to add a Github Action, but since we use Lerna to publish our packages, I think we have to do it this way. Instructions [found here](https://github.com/lerna/lerna/issues/3657#issuecomment-1535284598).

These are the changes:
* Lerna only added support for provenance statements in v6.6.2, so I had to update to this. As far as I can tell from running lerna commands locally, this seems to work.
* In order to update Lerna, I had to bump node by one minor version. The package's node version is still very much out of date and it would be nice to properly update this, but given that we might be looking at some bigger changes and updating Node requires rewriting a lot of our Storybook integration, I am just choosing to bump the minimum amount required. Storybook seems fine after this minor Node upgrade.
* Added the provenance option to the package.json

To tophat:
* Make sure our Lerna commands work
* Make sure Storybook works
* Once we ship, we will know if this has worked by checking at the bottom of the NPM page and seeing something like you can see here:

Screenshot 2025-02-25 at 3 48 21 PM

Part of https://github.com/Shopify/core-issues/issues/83848

N/A
 
## Storybook link
n/a

### Before merging


envex (approved):
Are the `yarn.lock` changes expected?

carysmills:
@envex they all appear to be related to the Lerna update, but let me know if anything sticks out to you

PR #1819, opened 2025-03-07
Funnel tooltip - explicitly set line height

michaelnesen:

We found a small discrepancy in https://github.com/Shopify/analytics-ui-components/pull/471 when viewing the funnel tooltips in web. Because we were not setting an explicit `line-height` in the Funnel tooltips, they were being inherited from web which was causing them to appear larger than desired. This PR just explicitly sets the `line-height` in the tooltip

## Storybook link
https://6062ad4a2d14cd0021539c1b-xtlgbpdvug.chromatic.com/?path=/story/polaris-viz-charts-funnelchartnext--with-trend-indicators

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
+++ b/packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
@@ -39,6 +39,7 @@ import {
   TOOLTIP_HEIGHT,
   TOOLTIP_WIDTH,
   TOOLTIP_RIGHT_POSITION_THRESHOLD,
+  TOOLTIP_BOTTOM_OFFSET,
 } from './constants';
 import {useBuildFunnelTrends} from './utilities/useBuildFunnelTrends';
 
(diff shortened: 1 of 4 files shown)


packages/polaris-viz/src/components/FunnelChartNext/Chart.tsx
```diff
@@ -286,7 +287,7 @@ export function Chart({
       return yPosition;
     }
 
-    return yPosition - TOOLTIP_HEIGHT;
+    return yPosition - TOOLTIP_HEIGHT - TOOLTIP_BOTTOM_OFFSET;
```
mollerjorge, line 290: Changing the line height tooltip rows slightly increased the size of the tooltip, however looking at the designs they are actually more inline now then they were before. I added this offset to make sure there is some space between the tooltip position and bar

mollerjorge (approved):

PR #1823, opened 2025-03-13
Add Support for Dynamic Annotation Popover Rendering

benwolfram:

In order to display custom annotations for an upcoming report, looking to provide support to render more than just a `string` within the annotations popover. Allowing for more dynamism and flexibility in how we represent data and annotations. This was the recommended approach discussing in Slack, and UX was also looking to gain this functionality.

Includes adding a `renderAnnotationContent` prop for all charts that support annotations (some don't seem to). I've split the commits so that the initial commit is adding the code for this, along with one usage for LineChart.

Relates to https://github.com/Shopify/temp-project-mover-Archetypically-20250318093910/issues/239

| Before  | After  |
|---|---|
|image|image|
 
## Storybook link

I've updated the Annotations story for LineChart.

http://localhost:6006/?path=/story/polaris-viz-charts-linechart--annotations

### Before merging

### DIFF
--- a/packages/polaris-viz/src/components/Annotations/components/AnnotationContent/AnnotationContent.tsx
+++ b/packages/polaris-viz/src/components/Annotations/components/AnnotationContent/AnnotationContent.tsx
@@ -1,10 +1,10 @@
-import {useEffect, useLayoutEffect, useState} from 'react';
+import {Fragment, useEffect, useLayoutEffect, useState} from 'react';
 import {createPortal} from 'react-dom';
 import {changeColorOpacity, clamp, useTheme} from '@shopify/polaris-viz-core';
 
 import {TOOLTIP_BG_OPACITY} from '../../../../constants';
 import {useBrowserCheck} from '../../../../hooks/useBrowserCheck';
-import type {Annotation} from '../../../../types';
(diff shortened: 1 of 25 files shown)


packages/polaris-viz/src/components/Annotations/components/AnnotationContent/AnnotationContent.tsx
```diff
@@ -84,47 +88,62 @@ export function AnnotationContent({
...
         id={`annotation-content-${index}`}
         role="dialog"
       >
         {title != null && (
           <p
             className={styles.Title}
-            style={{color: selectedTheme.tooltip.textColor}}
+            style={
+              {
+                '--selected-theme-tooltip-textColor':
```
envex, line 113: Can we get away from adding a variable here and then having to cast?

benwolfram:
[Force push notes](https://github.com/Shopify/polaris-viz/compare/a020de7d6d8c06a88887d17666dd4d9ec8b2067e..f7769733ff3f511b44c8ed1e0e3cd6d624984584)
- Remove CSS variable implementation, rely on inline style to pass colour

envex (approved):

PR #1825, opened 2025-03-26
Funnel chart null value handling

michaelnesen:

This PR's add's better `null` value handling to the funnel chart. Previously when calculating the dropped off - percentage change, we would treat `null` values as `0` which isn't correct and was an oversight when this logic was initially added. 

With these changes, if there are any null values passed into the chart we now gracefully handle them and show `-` trend in tooltips. We also don't show dropped off labels for segments next to null values (this wasn't being handled before).

No issue for this, related to: https://github.com/Shopify/analytics-ui-components/pull/471

Before (with null comparison value):

Screenshot 2025-03-26 at 4 53 55 PM

After (with null comparison value):
Screenshot 2025-03-26 at 4 55 28 PM

 
### Before merging


mollerjorge (approved):
Tested the changes and code looks good.

PR #1845, opened 2025-05-02
Render xAxis annotations above yAxisLine

envex:

We have to render one set of annotations above the other, so in some case an annotation line will render over the annotation tooltip. 

Until we can come up with a better solution to render the tooltips, we are going to prioritize rendering the xAxis annotations.

image

### Before merging


rita-morozova (approved):
🎩 'ing and code look good to me

PR #1851, opened 2025-05-14
Add new LineChartCumulative visualization

Passanelli:

Initial implementation for new `LineChartCumulative` visualization
Address this task: https://github.com/Shopify/merchant-analytics-issues/issues/617

For this first iteration this visualization will be rendered inside `HomeCards` so they are not interactive at the moment.

figma link: https://www.figma.com/design/9Sm7SuRkGTNTCgonxhXnWO/Push-insights-to-home-proposal?node-id=1105-5699&p=f&t=0qHDAS5ng3x6stIA-0

image
 
 
## Storybook link

- checkout current branch `passanelli/add_cumulative_line_chart`
- run `pnpm storybook`
- go to `Charts` -> `LineChartCumulative` 

### Before merging


packages/polaris-viz/src/components/LineChartCumulative/Chart.tsx
```diff
@@ -0,0 +1,199 @@
...
+            svgDimensions={{height: drawableHeight, width: drawableWidth}}
+            theme={theme}
+            xScale={xScale}
+            yScale={yScale}
+            type="default"
+            shouldShowArea={false}
+          />
+
+          {data.map((singleSeries, index) => {
+            if (singleSeries.metadata?.isVisuallyHidden === true) {
```
envex: Do you think we'll still need this? This is to enable consumers to hide lines.

packages/polaris-viz/src/components/LineChartCumulative/components/Points/Points.tsx
```diff
@@ -0,0 +1,81 @@
+import {Fragment} from 'react';
```
envex, line 1: Would we be able to re-use `Points` from `LineChart`, or do we need to duplicate the component?

packages/polaris-viz/src/components/LineChartCumulative/components/PointsAndCrosshair/PointsAndCrosshair.tsx
```diff
@@ -0,0 +1,68 @@
+import {Fragment, useRef} from 'react';
```
envex, line 1: Same here, can we reuse the existing component?

packages/polaris-viz/src/components/TextLine/TextLine.tsx
```diff
@@ -46,6 +47,7 @@ export function TextLine({color, index, line}: TextLineProps) {
                 fontSize={fontSize}
                 fontFamily={FONT_FAMILY}
                 transform={transform}
+                style={fontWeight ? {fontWeight} : {}}
```
envex: Can this be added to `FormattedLine`?

packages/polaris-viz/src/components/XAxis/XAxis.tsx
```diff
@@ -50,14 +55,20 @@ export function XAxis({
...
           <g
             transform={`translate(${x + (xPosition ?? 0)},${y})`}
             key={index}
             aria-hidden={ariaHidden}
           >
-            <TextLine line={line} index={index} />
+            <TextLine
+              line={line}
+              index={index}
+              color={isActiveIndex ? activeIndex.color : undefined}
```
envex: I wonder if we should move this to `FormattedLine` instead of adding a prop to `TextLine`.

.storybook/preview.js
```diff
@@ -163,6 +164,20 @@ export const decorators = [
               color: 'rgb(0, 164, 124)',
             },
           },
+          LinePositive: {
```
envex: Let's pass these colors along side the `DataSeries` instead of setting a new theme for each.

packages/polaris-viz/src/components/XAxis/XAxis.tsx
Passanelli: Actually the color prop was already there but we were not passing it any value from this `<XAxis />` component.. i can move it to `TextLine` instead if it's a better approach, just wanted to mention that

packages/polaris-viz/src/components/TextLine/TextLine.tsx
Passanelli: do you think that moving the `fontWeight` value to `FormattedLine` will be good or adding an `style` attr with the `fontWeight` value on it will be better? 🤔

envex: I think `style` is fine if `<text />` doesn't accept a `fontWeight` property, but I do think `fontWeight` can be added to `FortmattedLine`

packages/polaris-viz/src/components/XAxis/XAxis.tsx
envex: I think it probably makes sense to push color to `FormattedLine`.

The `activeIndex` being an object that includes a number and a color seems a little funky.

packages/polaris-viz/src/components/LineChartCumulative/Chart.tsx
Passanelli: oh i think it's safe to delete this yes

packages/polaris-viz/src/components/LineChartCumulative/components/Points/Points.tsx
Passanelli: I didn't re-use them to avoid to have some logic that will not be needed at least for this first instance of the visualization and this new components are quite simpler than the existing ones at `LineChart` level

packages/polaris-viz/src/components/LineChartCumulative/components/PointsAndCrosshair/PointsAndCrosshair.tsx
Passanelli: I didn't re-use them to avoid to have some logic that will not be needed at least for this first instance of the visualization and this new components are quite simpler than the existing ones at `LineChart` level

Passanelli:
will be handled at this other PR: https://github.com/Shopify/polaris-viz/pull/1855

